PR #148, opened 2013-01-18
Don't fail if the response of a HEAD request has a empty string encoded in gzip

abarre:
In our company, we have a http proxy in node.js that fails on a lot of HEAD response received from the customer servers.

After analysis, we found that some responses to HEAD requests don't have an empty body as they should have but always have 20 chars in the gzip content-type.

We discovered that the 20 chars are produced by gzipping an empty string.

```
-> % echo -n | gzip - > zip.txt && ls -l zip.txt && hexdump zip.txt
-rw-r--r--  1 anthony  wheel  20 18 jan 13:35 zip.txt
0000000 1f 8b 08 00 7a 41 f9 50 00 03 03 00 00 00 00 00
0000010 00 00 00 00                                    
0000014
```

In fact, it seems that this bug is from [nginx](http://trac.nginx.org/nginx/ticket/261).

For us, it's very annoying to fail (`HPE_INVALID_CONSTANT` at the `s_start_res` state) on this response parsing.

The patch will simply skip the analysis of the body response if the SKIPBODY is activated. It's more robust than continuing to parse the buffer even if the headers part if finished.


dax:
It seems the problem can be reproduced with any gzip response with `content-length: 0` header regardless of the HTTP verb.

PR #159, opened 2013-08-30
Add two hooks to be executed when a chunk is downloaded.

vincentbernat:
One hook is executed right before receiving a new chunk. The second
one is executed just after. parsertrace.c is updated to use those
hooks for display.

I am using this for experimentation with rendering. This is more reliable than looking at timestamps to know where chunks start and end.

### DIFF
--- a/contrib/parsertrace.c
+++ b/contrib/parsertrace.c
@@ -47,6 +47,18 @@ int on_message_complete(http_parser* _) {
   return 0;
 }
 
+int on_chunk_begin(http_parser* _) {
+  (void)_;
+  printf("\n***CHUNK BEGIN***\n\n");
+  return 0;
+}
+
+int on_chunk_complete(http_parser* _) {
+  (void)_;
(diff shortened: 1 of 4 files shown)


jasnell:
@vincentbernat ... is this still something you'd like to pursue? If so, the PR would need to be rebased and updated /cc @bnoordhuis @indutny

vincentbernat:
I could update the PR, but would it get more feedback today than it has two years ago? Same for #161 and #162. I understand there is no interest to merge and as I don't need those changes anymore, I don't want to spend more time on them.

jasnell:
@vincentbernat ... yeah, it would get more feedback as there are a lot more people looking now (That's no guarantee that it would land tho)

derekargueta:
Can this PR be closed since we have `on_chunk_header` and `on_chunk_complete` today?

bnoordhuis:
Thanks, it looks like this can indeed be closed.

PR #160, opened 2013-09-01
Clean up some typo & flow issues in README

dweekly:


jasnell:
Looks like most of these edits have already landed in master.

PR #161, opened 2013-09-02
Don't abort silently a test if a strict-mode check fails.

vincentbernat:
In `test_message_pause()`, we don't expect a strict-mode check to
fail. Don't abort silently the test if it is the case. Raise an assert
error instead. We keep a separate assert to help with debugging.

### DIFF
--- a/test.c
+++ b/test.c
@@ -3134,13 +3134,7 @@ test_message_pause (const struct message *msg)
     }
 
     if (nread < buflen) {
-
-      // Not much do to if we failed a strict-mode check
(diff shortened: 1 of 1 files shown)


jasnell:
@indutny ... I assume this one is safe to close given the lack of any activity on it. Can reopen if someone feels it is necessary.

vincentbernat:
The commit still applies and the problem is still here. I cannot reopen myself the PR and pushing a rebase has no effect.

jasnell:
I'll mark this one for a second look. It's not letting me re-open the PR either for some reason.

PR #184, opened 2014-08-05
Solaris doesn't necessarily have stdint.h, use inttypes.h

jacquesg:
Solaris doesn't necessarily have `stdint.h`, it's more portable to use `sys/inttypes.h`.

See:
http://wiki.opencsw.org/porting-faq#toc1

### DIFF
--- a/http_parser.h
+++ b/http_parser.h
@@ -42,6 +42,8 @@ typedef __int32 int32_t;
 typedef unsigned __int32 uint32_t;
 typedef __int64 int64_t;
 typedef unsigned __int64 uint64_t;
+#elif (defined(__sun) || defined(__sun__)) && defined(__SunOS_5_9)
+#include <sys/inttypes.h>
 #else
 #include <stdint.h>
 #endif


jasnell:
@indutny ... I assume this one is likely safe to close?

jacquesg:
Is there any reason this cannot be merged?

indutny:
@jasnell I have no idea, we clearly need input of Solaris people here.

jasnell:
hmm... I'm not a solaris person either :-/ ... @misterdjules ideas?

rmustacc:
On 12/26/16 12:36 , James M Snell wrote:
> hmm... I'm not a solaris person either :-/ ... @misterdjules ideas?
> 

In general, <stdint.h> is what user land software is supposed to include
and <sys/inttypes.h> is for device drivers. The problem seems to be that
older versions of Solaris (Solaris 9 and older) don't seem to have that
header. At the moment, it seems that the content from both <stdint.h>
and <sys/inttypes.h> is generally the same, though the set of
pre-processor macros being used could change it so you see less
definitions with only the latter.

I can't really speak for what modern Oracle Solaris has done, only
illumos, though it likely hasn't changed that much between the two.

The original desire to constrain this to older versions is probably so
that we're using the expected header on current systems and so that way
we don't end up in the case where there are additions that have been
made to the standard headers for user land that we somehow miss by
including a device driver header.

Hopefully this helps.

jacquesg:
I've constrained this to Solaris 9.

jacquesg:
Anything else required to get this merged?

derekargueta:
Is this PR still being considered?

jacquesg:
It is still relevant, atleast for me.

bnoordhuis:
I somehow managed to completely overlook this PR for nearly six years, sorry about that. Merged just now in 55e736c, thanks Jacques!

PR #213, opened 2015-01-14
src: fix build on MSVC

indutny:
Do explicit casts from/to enums, convert type of `memchr`'s return value
to `const char*`.

cc @piscisaureus

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -57,7 +57,7 @@ do {                                                                 \
 } while(0)
 
 #define CURRENT_STATE() p_state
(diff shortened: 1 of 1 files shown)


indutny:
See https://github.com/joyent/http-parser/issues/212

indutny:
Landed in 280af6984cb4d674e700854f40558832e6dcaa0b, thank you @piscisaureus!

piscisaureus:
You're welcome as always Fedor.

PR #214, opened 2015-01-20
Added support for SOURCE

ThisIsMissEm:
As per tweets: https://twitter.com/indutny/status/557499042243289088

It seems new handling for upgrade was added at some point, which makes the tests break. Unfortunately I don't have time to work out how to make the tests pass again.

SOURCE should essentially be handled like Upgrade, I believe, and switched to data/tcp mode after parsing headers.

/cc @indutny @bnoordhuis

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -969,7 +969,7 @@ size_t http_parser_execute (http_parser *parser,
             /* or PROPFIND|PROPPATCH|PUT|PATCH|PURGE */
             break;
           case 'R': parser->method = HTTP_REPORT; break;
-          case 'S': parser->method = HTTP_SUBSCRIBE; /* or SEARCH */ break;
+          case 'S': parser->method = HTTP_SUBSCRIBE; /* or SEARCH, SOURCE */ break;
           case 'T': parser->method = HTTP_TRACE; break;
           case 'U': parser->method = HTTP_UNLOCK; /* or UNSUBSCRIBE */ break;
           default:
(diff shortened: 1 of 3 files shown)


indutny:
I'm quite sure that it doesn't affect it, but nevertheless.

May I ask you to run the benchmark before and after the change and post the results here?

miksago:
This changeset would also affect the decision of the Icecast project to move to using http-parser: https://trac.xiph.org/ticket/2057

miksago:
With patch:

```
Benchmark result:
Took 12.565303 seconds to run
397921.156250 req/sec
```

Without:

```
Benchmark result:
Took 14.347905 seconds to run
348482.937500 req/sec
```

I'm not particularly sure why it took longer without the patch. Ran a second time, and got:

```
Benchmark result:
Took 11.664329 seconds to run
428657.343750 req/sec
```

indutny:
Ok, I see. Are tests passing on your machine? It seems that CI is not happy with it.

miksago:
My machine is under pretty heavy load (memory, disk) and may have a hard drive failure happening soon.

miksago:
No, see my note in the comment about the fix_upgrade_message failing. I don't know how to fix straight away, and can't put any more time in right now

jasnell:
@indutny ... any further thoughts on this one?

jasnell:
I will note that `SOURCE` is not in the official list of registered methods: http://www.iana.org/assignments/http-methods/http-methods.xhtml

miksago:
Indeed, it is not, although, it also turns out the icecast protocol isn't exactly HTTP anyway. 

The reason for this change was a suggestion to the icecast dev-team to try using this http 
parser instead of rewriting their own.

jasnell:
While that kind of reuse is definitely a noble goal, I'm not sure it's the right thing to do in http-parser right now. The [HTTP WG](http://github.com/nodejs/http) will be taking another look at HTTP support in core. Part of that will be revisiting the http-parser and exploring what additional changes should be made -- including support for arbitrary HTTP methods as opposed to a limited fixed list. Up to this point, the philosophy has been to support only methods that are included in the official registry.

ThisIsMissEm:
@jasnell any updates needed here?

ThisIsMissEm:
Looks like this may have been fixed by #431 and #407, though those didn't treat the request as being a switching protocols type request, so, no upgrade event.

PR #215, opened 2015-01-23
Add function to initialize http_parser_settings

RomainGiraud:
When http_parser_settings struct is defined in function or class scope, struct members are not initilized to 0.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -2134,6 +2134,19 @@ http_parser_init (http_parser *parser, enum http_parser_type t)
   parser->http_errno = HPE_OK;
 }
 
+void
+http_parser_settings_init(http_parser_settings *settings)
+{
+  settings->on_message_begin = 0;
+  settings->on_url = 0;
+  settings->on_status = 0;
+  settings->on_header_field = 0;
+  settings->on_header_value = 0;
+  settings->on_headers_complete = 0;
+  settings->on_body = 0;
+  settings->on_message_complete = 0;
+}
+
 const char *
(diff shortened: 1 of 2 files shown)


http_parser.c
```diff
@@ -2134,6 +2134,19 @@ http_parser_init (http_parser *parser, enum http_parser_type t)
   parser->http_errno = HPE_OK;
 }
 
+void
+http_parser_settings_init(http_parser_settings *settings)
```
indutny: Style `void http_parser...(...) {`

http_parser.c
```diff
@@ -2134,6 +2134,19 @@ http_parser_init (http_parser *parser, enum http_parser_type t)
...
+{
+  settings->on_message_begin = 0;
```
indutny: Why not just `memset()`? ;)

marcomorain: Yeah, `memset` will be more robust if members are changed / added to the
struct.

http_parser.c
RomainGiraud: All function definitions have this same style.

http_parser.c
RomainGiraud: Indeed, I will change that.

http_parser.c
indutny: Arrrgh, you are right! Nginx legacy, damn it.

indutny:
LGTM, thanks.

indutny:
Landed in 53063b7, thank you!

PR #216, opened 2015-01-24
src: skip lws between `connection` values

indutny:
Fix https://github.com/iojs/io.js/issues/588

cc @bnoordhuis

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -1618,6 +1618,8 @@ size_t http_parser_execute (http_parser *parser,
                 h_state = h_matching_connection_upgrade;
               } else if (STRICT_TOKEN(c)) {
                 h_state = h_matching_connection_token;
+              } else if (c == ' ' || c == '\t') {
+                /* Skip lws */
               } else {
                 h_state = h_general;
               }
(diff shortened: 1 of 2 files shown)


bnoordhuis:
LGTM with the nit that RFC 2616 allows multi-line continuations like `Upgrade: \n keep-alive`.  Perhaps you can address that in a follow-up PR, I'll leave that up to you.  This PR is already strictly an improvement, no reason not to land it.

indutny:
I thought this behavior was deprecated in the last http 1.1 rfc, wasn't it?

bnoordhuis:
I think - more correctly: I fear - there is no such thing as 'deprecated' when it comes to popular protocols.

indutny:
@bnoordhuis anyway, it should be working, AFAIK. Take a look at the test above ^

indutny:
@bnoordhuis : ping ;)

bnoordhuis:
@indutny It still doesn't handle the `Upgrade: \n keep-alive` case, does it?  But like I said, it's already an improvement, so LGTM.

indutny:
Landed in 59569f2125211ed7ce4163f34d0cc6f4b6801f1c, thank you!

PR #219, opened 2015-02-15
header: fix field sizes

### LINKED ISSUE #218: on_headers_complete handling is broken
Returning '1' from on_headers_complete is supposed to skip body processing, but it doesn't, since "F_SKIPBODY = 1 << 6" doesn't fit into 6-bit "unsigned int flags:6".

indutny:
Fix #218

cc @bnoordhuis

### DIFF
--- a/http_parser.h
+++ b/http_parser.h
@@ -204,8 +204,8 @@ enum http_errno {
 struct http_parser {
   /** PRIVATE **/
   unsigned int type : 2;         /* enum http_parser_type */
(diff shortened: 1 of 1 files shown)


bnoordhuis:
LGTM but maybe say 'use unused bits' instead of 'fix' in the commit log.

indutny:
@bnoordhuis take a look at #218, flags do not fit the field anymore.

bnoordhuis:
Alright then, can you update the commit log with an explanation of the fix? :-)

indutny:
Landed in 7ba3123, thank you!

Issue #220, opened 2015-02-19
parser->nread incorrect after removing reexecute goto


jpinner:
Commit 959f4cb replaced the "reexecute_byte" goto with a pointer decrement and break statement. The result is that whenever a byte is re-executed, the count header size macro is also re-executed, resulting in an incorrect value for parser->nread.

indutny:
Confirmed. Looking.

indutny:
Should be fixed by https://github.com/joyent/http-parser/pull/221

indutny:
Fixed in 7ecf775, thank you!

jpinner:
one nit: unnecessary trailing '\' in the #define

PR #221, opened 2015-02-19
src: partially revert 959f4cb to fix nread value

indutny:
With 959f4cb on reexecution - header byte was accounted twice, leading
to the incorrect `parser->nread` value.

Fix: #220


indutny:
cc @bnoordhuis

bnoordhuis:
LGTM

indutny:
Thank you, landed in 7ecf775 !

PR #222, opened 2015-02-23
src: fix url parsing crash on missing host

jay:
(http_parse_host)
- Check for invalid parameters.
- Cast field offset assignment to proper type.

(http_parser_parse_url)
- Check for invalid parameters.
- Cast field offset assignment to proper type.
- Fail if the schema field was found but the host field was not found.
- Call http_parse_host only if the host field was found.

Prior to this change http_parse_host, which expects the UF_HOST field to
be filled, could be called even when that field was not filled resulting
in a crash. Refer to issue joyent/http-parser#209.


indutny:
Thanks, but this change is unnecessary. The problem is that the structure should be zero-initialized and this is not documented anywhere (nor used in url_parser).

PR #223, opened 2015-02-23
src: fix msvc warning on while(0) in macros

jay:
- Add a WHILE_FALSE macro to handle while(0) so that the msvc warning
"C4127: conditional expression is constant" can be suppressed in
function-like macros.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -29,6 +29,20 @@
 #include <string.h>
 #include <limits.h>
 
+#define WHILE_FALSE while(0)
+#if defined(_MSC_VER) && !defined(__POCC__)
+#  undef WHILE_FALSE
+#  if (_MSC_VER < 1500)
+#    define WHILE_FALSE while(1, 0)
+#  else
+#    define WHILE_FALSE \
+__pragma(warning(push)) \
+__pragma(warning(disable:4127)) \
+while(0) \
+__pragma(warning(pop))
+#  endif
+#endif
+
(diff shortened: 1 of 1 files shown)


http_parser.c
```diff
@@ -29,6 +29,20 @@
 #include <string.h>
 #include <limits.h>
 
+#define WHILE_FALSE while(0)
+#if defined(_MSC_VER) && !defined(__POCC__)
+#  undef WHILE_FALSE
+#  if (_MSC_VER < 1500)
+#    define WHILE_FALSE while(1, 0)
+#  else
+#    define WHILE_FALSE \
```
indutny: Style, please line up `\` as we do in defines below.

indutny:
I don't have much objections, cc @piscisaureus @bnoordhuis

bnoordhuis:
Please god, no.  I'm all for warning free builds but MS should just fix their damn compiler and make it understand this extremely common C idiom.

indutny:
Ahahah :) Thanks, Ben.

piscisaureus:
I have never seen this warning and I think it's useless. Maybe just disable the warning in `http_parser.gyp`? Add a directive like this: https://github.com/joyent/node/blob/885f72105bf0b18007100550f131c85ca56536e0/common.gypi#L154, but specifying 4127 as the disabled warning.

indutny:
Guess we should close this then. Thanks!

PR #224, opened 2015-02-24
url_parser: remove mixed declarations

jay:
(dump_url)
- Remove erroneous quote in output.

(main)
- Remove mixed declarations for compatibility with pre-c99 compilers
like msvc.


indutny:
LGTM

indutny:
Landed in d547f3b, thank you!

PR #225, opened 2015-02-24
src: introduce `http_parser_url_init`

### LINKED ISSUE #209: http_parser_parse_url crash when random string passed
was expecting to get only path UF_PATH retrieved when calling http_parser_parse_url without any schema or "//", instead the function crashes the host app.

it should either fail to fill any of the fields or just fill in path with full string if no schema "<schema>//.." specified

indutny:
The struct must be zero-initialized, but this wasn't explicitly stated
anywhere in headers. Introduce `http_parser_url_init` API method that
will do it.

Fix #209 

cc @bnoordhuis

### DIFF
--- a/contrib/url_parser.c
+++ b/contrib/url_parser.c
@@ -35,6 +35,7 @@ int main(int argc, char ** argv) {
   connect = strcmp("connect", argv[1]) == 0 ? 1 : 0;
   printf("Parsing %s, connect %d\n", argv[2], connect);
 
+  http_parser_url_init(&u);
   result = http_parser_parse_url(argv[2], len, connect, &u);
   if (result != 0) {
     printf("Parse error : %d\n", result);
--- a/http_parser.c
+++ b/http_parser.c
@@ -2292,6 +2292,11 @@ http_parse_host(const char * buf, struct http_parser_url *u, int found_at) {
   return 0;
 }
 
+void
+http_parser_url_init(struct http_parser_url *u) {
+  memset(u, 0, sizeof(*u));
+}
+
 int
 http_parser_parse_url(const char *buf, size_t buflen, int is_connect,
                       struct http_parser_url *u)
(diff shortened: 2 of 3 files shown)


jay:
My pr at #222 is more comprehensive, I'm not sure why you closed it as unnecessary. I added the parameter checking and zeroed the structure. In `http_parse_host` there's no point continuing without the host field and in `http_parser_parse_url` there's no point calling the former without the host field. Also something should be done to make sure that the buflen isn't greater than max `uint16_t`, which I covered in both.

Also it occurs to me now an overflow check is also due for `size_t buflen` assignment in `http_parse_host` since there's no guarantee `size_t` is greater than `uint16_t` (although it's extremely likely), which would probably look something like this, placed immediately after the assignment:
```c
if(u->field_data[UF_HOST].off < buflen) { // overflow
    return 1;
}
```

indutny:
It won't call it without host field, because UF_HOST won't be set. In case of `test` string it is doing a right thing and treating `test` as a schema, and then awaits for the hostname.

jay:
In the case a schema field was found but a host field was not then `http_parser_parse_url` will still call `http_parse_host`. I changed that by failing if a schema field is found without a host field and calling `http_parse_host` only if a host field was found as seen [here](https://github.com/joyent/http-parser/pull/222/files#diff-5c62f371bf37583234d2462ad49ce33dL2366).

indutny:
Right and `http_parse_host` is called because it is what is expected after the scheme. It is intentionally invoked for this thing to fail. Practically, is there any difference in behavior your PR and current implementation (in master branch)?

jay:
Yes it's different. The current implementation is calling `http_parse_host` when `UF_HOST` is not set which contradicts what you said earlier. Whether that's correct or not you would know having more familiarity with the code base, however I did not think that was correct.

indutny:
Right, I was wrong about it. Sorry. However, except this I don't see any *behavior* difference.

indutny:
@mscdex may I ask you to review it? ;)

jasnell:
@indutny @mscdex ... any further thoughts on this one?

mscdex:
LGTM

jasnell:
Landed in 777ba4e

PR #228, opened 2015-03-12
install/uninstall targets in Makefile

Umorrian:
For easier installation as shared library.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,7 +113,7 @@ install: library
 	install -D $(SONAME) $(LIBDIR)/$(SONAME)
 	-mv http_parser.h $(INCLUDEDIR)/http_parser.h
  
-install-strip: release
+install-strip: library
 	install -D -s $(SONAME) $(LIBDIR)/$(SONAME)
 	-mv http_parser.h $(INCLUDEDIR)/http_parser.h
  
@@ -129,4 +129,4 @@ clean:
 contrib/url_parser.c:	http_parser.h
 contrib/parsertrace.c:	http_parser.h
 
-.PHONY: clean package test-run test-run-timed test-valgrind install
+.PHONY: clean package test-run test-run-timed test-valgrind install install-strip uninstall
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,7 @@ CFLAGS_LIB = $(CFLAGS_FAST) -fPIC
 
 LDFLAGS_LIB = $(LDFLAGS) -shared
 
+INSTALL ?= install
 PREFIX ?= $(DESTDIR)/usr/local
 LIBDIR = $(PREFIX)/lib
 INCLUDEDIR = $(PREFIX)/include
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -42,6 +42,10 @@ CFLAGS_LIB = $(CFLAGS_FAST) -fPIC
 
 LDFLAGS_LIB = $(LDFLAGS) -shared
 
+PREFIX = $(DESTDIR)/usr/local
```
indutny: Perhaps, it could be made overridable? Like `PREFIX ?= ...`

indutny:
LGTM, but I bet @bnoordhuis might have something to say. I'm totally dumb about file structure of library installations on unixes.

Makefile
```diff
@@ -105,6 +109,21 @@ parsertrace_g: http_parser_g.o contrib/parsertrace.c
 tags: http_parser.c http_parser.h test.c
 	ctags $^
 
+install: library
+	install -D $(SONAME) $(LIBDIR)/$(SONAME)
```
bnoordhuis: Can you add `INSTALL ?= install` at the top and use that instead of a literal `install`?

Makefile
```diff
@@ -105,6 +109,21 @@ parsertrace_g: http_parser_g.o contrib/parsertrace.c
...
+	ln -s $(LIBDIR)/$(SONAME) $(LIBDIR)/libhttp_parser.so
```
bnoordhuis: Maybe place this after the install of the header so that erroring on an existing symlink doesn't leave a partially installed libhttp_parser (i.e. the library gets installed but not the headers.)

bnoordhuis:
LGTM with suggestions.

Umorrian:
Implemented @bnoordhuis suggestions. I otherwise never write actual Makefiles, just got annoyed doing the installation manually.

bnoordhuis:
LGTM

indutny:
Landed in 5d414fc, thank you!

Makefile
```diff
@@ -110,16 +110,19 @@ tags: http_parser.c http_parser.h test.c
...
+        install -D $(SONAME) $(LIBDIR)/$(SONAME)
+        ln -s $(LIBDIR)/$(SONAME) $(LIBDIR)/libhttp_parser.so
+        install -D  http_parser.h $(INCLUDEDIR)/http_parser.h
+
 install-strip: library
-	install -D -s $(SONAME) $(LIBDIR)/$(SONAME)
-	-mv http_parser.h $(INCLUDEDIR)/http_parser.h
- 
+        install -D -s $(SONAME) $(LIBDIR)/$(SONAME)
+        ln -s $(LIBDIR)/$(SONAME) $(LIBDIR)/libhttp_parser.so
```
eli-schwartz: This is incredibly wrong and utterly violates the fundamental aspect of DESTDIR. See https://www.gnu.org/prep/standards/html_node/DESTDIR.html

Also see #264 #272 #321 and #346 which all try to fix this and thereby allow the Makefile to actually work with DESTDIR, all of which are busy being ignored.

PR #233, opened 2015-04-18
Closes #231 : add callbacks for processing chunk boundaries

### LINKED ISSUE #231: add callbacks for processing chunk boundaries: on_chunk_header & on_chunk_complete
Some background first: the Proxygen team forked http_parser around 2011 (see https://github.com/facebook/proxygen/blob/master/proxygen/external/http_parser/http_parser.h) and now would like to get back to using regular http_parser releases, undoing the fork, pulling http_parser as a regular external dependency. Of all the changes required for undoing the Proxygen fork this one here is the only one that requires making an API change in http_parser (adding 2 callbacks). You think adding these 2 callbacks will be acceptable? This change here allowed Proxygen to maintained the original chunking when proxying requests. I figured I create a Github issue before bothering to port this feature back to the master repo & creating a pull request.

This here is probably a rehash of abandoned(?) issue #9.

KjellSchubert:
on_chunk_header & on_chunk_complete (Proxygen fork merge D508755 D521404, orig author simpkins@fb.com). Please see #231 for more details.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -2147,13 +2147,15 @@ http_parser_settings_init(http_parser_settings *settings)
 
 const char *
 http_errno_name(enum http_errno err) {
-  assert(((size_t)err) < (sizeof(http_strerror_tab)/sizeof(http_strerror_tab[0])));
+  assert(((size_t)err)
+      < (sizeof(http_strerror_tab)/sizeof(http_strerror_tab[0])));
   return http_strerror_tab[err].name;
 }
 
 const char *
 http_errno_description(enum http_errno err) {
-  assert(((size_t)err) < (sizeof(http_strerror_tab)/sizeof(http_strerror_tab[0])));
+  assert(((size_t)err)
+      < (sizeof(http_strerror_tab)/sizeof(http_strerror_tab[0])));
   return http_strerror_tab[err].description;
 }
 
--- a/http_parser.c
+++ b/http_parser.c
@@ -123,7 +123,7 @@ do {                                                                 \
     FOR##_mark = NULL;                                               \
   }                                                                  \
 } while (0)
-
+  
 /* Run the data callback FOR and consume the current byte */
 #define CALLBACK_DATA(FOR)                                           \
     CALLBACK_DATA_(FOR, p - FOR##_mark, p - data + 1)
--- a/http_parser.c
+++ b/http_parser.c
@@ -2147,13 +2147,13 @@ http_parser_settings_init(http_parser_settings *settings)
 
 const char *
 http_errno_name(enum http_errno err) {
-  assert(err < (sizeof(http_strerror_tab)/sizeof(http_strerror_tab[0])));
+  assert(((size_t)err) < (sizeof(http_strerror_tab)/sizeof(http_strerror_tab[0])));
   return http_strerror_tab[err].name;
 }
 
 const char *
 http_errno_description(enum http_errno err) {
-  assert(err < (sizeof(http_strerror_tab)/sizeof(http_strerror_tab[0])));
+  assert(((size_t)err) < (sizeof(http_strerror_tab)/sizeof(http_strerror_tab[0])));
   return http_strerror_tab[err].description;
 }
 
--- a/http_parser.c
+++ b/http_parser.c
@@ -123,7 +123,7 @@ do {                                                                 \
     FOR##_mark = NULL;                                               \
   }                                                                  \
 } while (0)
-  
+
 /* Run the data callback FOR and consume the current byte */
 #define CALLBACK_DATA(FOR)                                           \
     CALLBACK_DATA_(FOR, p - FOR##_mark, p - data + 1)
@@ -1782,9 +1782,9 @@ reexecute:
 
         if (parser->flags & F_TRAILING) {
           /* End of a chunked request */
-          UPDATE_STATE(NEW_MESSAGE());
-          CALLBACK_NOTIFY(message_complete);
-          break;
+          UPDATE_STATE(s_message_done);
+          CALLBACK_NOTIFY_NOADVANCE(chunk_complete);
+          REEXECUTE();
         }
 
         UPDATE_STATE(s_headers_done);
@@ -1991,8 +1991,10 @@ reexecute:
         if (parser->content_length == 0) {
           parser->flags |= F_TRAILING;
           UPDATE_STATE(s_header_field_start);
+          CALLBACK_NOTIFY(chunk_header);
         } else {
           UPDATE_STATE(s_chunk_data);
+          CALLBACK_NOTIFY(chunk_header);
         }
         break;
       }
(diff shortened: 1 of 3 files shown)


KjellSchubert:
Fails to compile with clang 3.5, I'll look into this (probably only on Monday).

http_parser.c
```diff
@@ -123,7 +123,7 @@ do {                                                                 \
     FOR##_mark = NULL;                                               \
   }                                                                  \
 } while (0)
-  
+
```
indutny: Unnecessary style fix.

http_parser.c
```diff
@@ -1782,9 +1782,9 @@ size_t http_parser_execute (http_parser *parser,
 
         if (parser->flags & F_TRAILING) {
           /* End of a chunked request */
-          UPDATE_STATE(NEW_MESSAGE());
-          CALLBACK_NOTIFY(message_complete);
-          break;
+          UPDATE_STATE(s_message_done);
```
indutny: I know it does the same thing internally, but why is it necessary to change it? Just for consistency?

test.c
```diff
@@ -1839,6 +1866,32 @@ response_status_cb (http_parser *p, const char *buf, size_t len)
   return 0;
 }
 
+int
+chunk_header_cb (http_parser *p)
+{
+  assert(p == parser);
+  if (p->content_length == 0) {
+    // Terminating chunk.  Don't increment num_chunks in this case.
```
indutny: Why?

test.c
```diff
@@ -3488,7 +3596,12 @@ main (void)
         , { "Content-Type", "text/plain" }
         }
       ,.body_size= 31337*1024
+      ,.num_chunks= 31337
+      ,.num_chunks_complete= 31338
       };
+    for (i = 0; i < MAX_CHUNKS; ++i) {
```
indutny: I'd rather use `i++` throughout this file for consistency.

test.c
```diff
@@ -1839,6 +1866,32 @@ response_status_cb (http_parser *p, const char *buf, size_t len)
...
+  }
+
+  return 0;
+}
+
+int
+chunk_complete_cb (http_parser *p)
+{
+  assert(p == parser);
+  ++messages[num_messages].num_chunks_complete;
```
indutny: Let's use postfix `++` here.

test.c
```diff
@@ -1839,6 +1866,32 @@ response_status_cb (http_parser *p, const char *buf, size_t len)
...
+chunk_header_cb (http_parser *p)
+{
+  assert(p == parser);
+  if (p->content_length == 0) {
+    // Terminating chunk.  Don't increment num_chunks in this case.
+    return 0;
+  }
+
+  int chunk_idx = messages[num_messages].num_chunks;
+  ++messages[num_messages].num_chunks;
```
indutny: Ditto.

indutny:
Some questions, otherwise LGTM! Thank you!

http_parser.c
KjellSchubert: I'm not the original author (just doing the selective merge), but I figure this change here is necessary to get the chunk_complete event out at the right time. Are you suggesting an alternative simpler solution? I can't think of one.

indutny: Well, the previous code has quite similar behaviour, except that the state will be different in chunk_complete_cb. I just want to be sure I understand the reasons behind this. Is it possible to ask the person who implemented it?

KjellSchubert: Sure, I asked him (by mail) to respond to some of your questions earlier this morning, he'll hopefully chime in soon. In the mean time I tried an alternative implementation that;s slightly closer to your original one:
```
         UPDATE_STATE(s_message_done);
          CALLBACK_NOTIFY_NOADVANCE(chunk_complete);
          UPDATE_STATE(NEW_MESSAGE());
          CALLBACK_NOTIFY(message_complete);
          break;
```
This variant passes tests and just adds the 2 lines for firing the chunk_complete event, so the diff may look slightly less scary. I have no strong preference for either solution, just slightly prefer the shorter committed one that reuses the 'case s_message_done:' (fewer LOC, plus we had this variant in production in the Proxygen branch for a long time).

test.c
KjellSchubert: I added some tracing for processing the test.c:TWO_CHUNKS_MULT_ZERO_END message, here the output:

```
CALLBACK_NOTIFY_ message_begin
CALLBACK_DATA_ url mark[0]=/ len=25
CALLBACK_DATA_ header_field mark[0]=T len=17
CALLBACK_DATA_ header_value mark[0]=c len=7
CALLBACK_NOTIFY_ chunk_header
test.c:chunk_header_cb p->content_length=5
CALLBACK_DATA_ body mark[0]=h len=5
CALLBACK_NOTIFY_ chunk_complete
CALLBACK_NOTIFY_ chunk_header
test.c:chunk_header_cb p->content_length=6
CALLBACK_DATA_ body mark[0]=  len=6
CALLBACK_NOTIFY_ chunk_complete
CALLBACK_NOTIFY_ chunk_header
test.c:chunk_header_cb p->content_length=0
CALLBACK_NOTIFY_ chunk_complete
CALLBACK_NOTIFY_ message_complete
```

I hope this answers your question: the chunk_header_cb is fired even for the final len=0 (pseudo) chunk, and the tests validate the number of chunks excluding this final len=0 one. Here you could ask: do we really want chunk_header + chunk_complete events for this final length=0 chunk or not. I prefer having this callback (since the parser's job is to markup byte blocks in the byte stream with their meaning, including the meaning for 0 or 000 as a chunk list terminator).

indutny: Yep, it just feels a bit odd that you have different number of complete and seen chunks in test declarations :)

KjellSchubert: True, I didn't understand this part fully either. In the tests I see stuff like
```
,.num_chunks= 0
,.num_chunks_complete= 1
```
but it looks to me like num_chunks_complete is always one larger than num_chunks, so whats the point of having both? Let me see if I can collapse these to one, checking with the original proxygen changeset's author if necessary...

KjellSchubert: OK, removed the redundant num_chunks in test declarations, see 33a79079b56af1390398d7862d2dba865e2e0002.

http_parser.c
simpkins: The previous code didn't call the chunk_complete at this location at all.  Setting the state to s_message_done before calling the chunk_complete callback is necessary for correctness in case the chunk_complete pauses the parser before returning.  This ensures that the state will be s_message_done on the next execution even if CALLBACK_NOTIFY_NOADVANCE(chunk_complete) ends up returning.  This makes sure that the message_complete callback will still be called correctly after the chunk_complete callback.

If you are asking for the overall motivation for adding these chunking callback, this helps implement proxies more efficiently.  Even though the chunk bounadaries shouldn't have semantic meaning, it is more efficient for the proxy to retransmit the message using the exact same chunk boundaries as originally used in the message.  Otherwise it is pessimistically forced to rechunk with smaller chunk boundaries in some cases, resulting in more data transmitted on the wire (due to more chunk headers).

indutny: Right! Thank you!

indutny:
Landed with minor fixes in d767545, thank you!

PR #235, opened 2015-04-20
Closes #234: No immediate onMessageComplete cb for upgrade w/body (D1364...

### LINKED ISSUE #234: incorrect handling of upgrade header for messages with non-empty body (e.g. POST)
This is one of the bugfixes in the Proxygen fork of http_parser: http_parser handling of the Upgrade header isn't really correct. If it sees an Upgrade header it immediately gives an onMessageComplete callback, sets the parser state to a new message and returns. For messages with an entity body, the upgrade wouldn't take effect until after the body was sent.

KjellSchubert:
...677 + D1380182 orig author afrind@fb.com)

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -123,7 +123,7 @@ do {                                                                 \
     FOR##_mark = NULL;                                               \
   }                                                                  \
 } while (0)
-
+  
 /* Run the data callback FOR and consume the current byte */
 #define CALLBACK_DATA(FOR)                                           \
     CALLBACK_DATA_(FOR, p - FOR##_mark, p - data + 1)
--- a/http_parser.c
+++ b/http_parser.c
@@ -1832,7 +1832,8 @@ reexecute:
 
         parser->nread = 0;
 
-        int hasBody = parser->flags & F_CHUNKED || parser->content_length > 0;
+        int hasBody = parser->flags & F_CHUNKED ||
+          (parser->content_length > 0 && parser->content_length != ULLONG_MAX);
         if (parser->upgrade && (parser->method == HTTP_CONNECT ||
                                 (parser->flags & F_SKIPBODY) || !hasBody)) {
           /* Exit, the rest of the message is in a different protocol. */
--- a/http_parser.c
+++ b/http_parser.c
@@ -123,7 +123,7 @@ do {                                                                 \
     FOR##_mark = NULL;                                               \
   }                                                                  \
 } while (0)
-  
+
 /* Run the data callback FOR and consume the current byte */
 #define CALLBACK_DATA(FOR)                                           \
     CALLBACK_DATA_(FOR, p - FOR##_mark, p - data + 1)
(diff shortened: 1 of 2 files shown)


indutny:
Very interesting! May I ask you to paste a quote from the RFC that allows POST+Upgrade+body? I just want to be sure that we don't introduce any non-standard behavior by this ;)

KjellSchubert:
>May I ask you to paste a quote from the RFC that allows POST+Upgrade+body?

I wasn't sure if you were kidding, in any case I checked https://tools.ietf.org/html/rfc7230#section-6.7 &  https://tools.ietf.org/html/rfc7231#section-6.2.2 : these talk about upgrade headers but I see no wording that limits this upgrade header to certain kinds of requests: like GET only but not POST, or requests with empty body but not non-empty body. In fact the word 'body' isnt mentioned around the Upgrade headers RFC section. One RFC quote "For example, if the Upgrade header field is received in a GET request and the server decides to switch protocols ..." - this implies to me that the Upgrade header can be contained in other requests but GETs, so why not POSTs with body for example? But in any case, this is still natural language, no one can prove or disprove anything with that :)

You honestly think Upgrade header with HTTP POST (or other HTTP verb + body) could be illegal? If so please quote an RFC that disallows that ;) This would be a weird un-orthogonal protocol limitation imo.

indutny:
@KjellSchubert thank you so much! ( I wasn't kidding, btw :) )

It seems that RFC allows it indeed.

LGTM then. @mscdex: may I ask you to take a look at this too? You seem to have lots of insights into HTTP spec.

mscdex:
@indutny As far as I can tell from RFC 7230/7231, it appears to be valid. Although I suspect it's about as common as sending a GET request with a body :-)

indutny:
Landed in dff604d, thank you!

samoconnor:
Hi @mscdex @KjellSchubert @indutny,
I'm a bit late to the party here, but I'd like your opinions on the `CONNECT_WITH_BODY_REQUEST` test case. The input has `Content-Length: 10` and a body `"blarfcicle"`, but the test expects [`.body= ""`](https://github.com/nodejs/http-parser/blob/master/test.c#L1102) !

[The RFC says](https://tools.ietf.org/html/rfc7230#section-3.3):
> _The presence of a message body in a request is signaled by a
    Content-Length or Transfer-Encoding header field.  Request message
   framing is independent of method semantics, even if the method does
   not define any use for a message body._

i.e. A request that says `Content-Length: 10` has a 10-byte body irrespective of the method being `CONNECT`.

The `CONNECT` scemantics are defined here: [RFC 7231, 4.3.6](https://tools.ietf.org/html/rfc7231#section-4.3.6). It talks about the client ignoring *responses( with Content-Length in, however, that is not relevant to this test *request* test case.

>    _A server MUST NOT send any Transfer-Encoding or Content-Length header
   fields in a 2xx (Successful) response to CONNECT.  A client MUST
   ignore any Content-Length or Transfer-Encoding header fields received
   in a successful response to CONNECT._

[RFC 7231, 4.3.6](https://tools.ietf.org/html/rfc7231#section-4.3.6) goes on to say that a payload on a `CONNECT` request has no defined meaning, but it does not disallow it:

>  _A payload within a CONNECT request message has no defined semantics;
   sending a payload body on a CONNECT request might cause some existing
   implementations to reject the request._

So, this line may be wrong: https://github.com/nodejs/http-parser/pull/235/files#diff-5c62f371bf37583234d2462ad49ce33dR1837

(Background: @quinnj has ported the node http parser to Julia, which means we've inherited the test cases, and I've been testing it and reviewing it vs the RFC).

PR #237, opened 2015-04-21
src: handle `Upgrade:` without `Conn: upgrade`

indutny:
In strict mode this should not be treated as upgrade, in normal mode we
should be gentle and allow it.

Fix: https://github.com/joyent/http-parser/issues/217

cc @bnoordhuis @mscdex - this is a compromise solution, let's continue discussion here.


indutny:
Decided to not go this awful way for now :)

PR #246, opened 2015-05-18
Fix invalid memory access.

OnixGH:
http_parse_host() depends on u->field_data[UF_HOST], but this if()
allowed the method to be called if only u->field_data[UF_SCHEMA] was
set, resulting in use of unintialized pointers.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -2376,7 +2376,7 @@ http_parser_parse_url(const char *buf, size_t buflen, int is_connect,
 
   /* host must be present if there is a schema */
   /* parsing http:///toto will fail */
-  if ((u->field_set & ((1 << UF_SCHEMA) | (1 << UF_HOST))) != 0) {
+  if ((u->field_set & (1 << UF_SCHEMA)) != 0 && (u->field_set & (1 << UF_HOST)) != 0) {
     if (http_parse_host(buf, u, found_at) != 0) {
       return 1;
     }
--- a/http_parser.c
+++ b/http_parser.c
@@ -2376,12 +2376,8 @@ http_parser_parse_url(const char *buf, size_t buflen, int is_connect,
 
   /* host must be present if there is a schema */
   /* parsing http:///toto will fail */
-  if ((u->field_set & (1 << UF_SCHEMA)) && (u->field_set & (1 << UF_HOST)) == 0) {
-    return 1;
-  }
-
-  /* CONNECT requests can only contain "hostname:port" */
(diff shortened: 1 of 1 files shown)


OnixGH:
This patch fixed a recurring crash in Phusion Passenger [(issue report)](https://github.com/phusion/passenger/issues/1513).

OnixGH:
The thread for [225](https://github.com/joyent/http-parser/pull/225) is related. However, it seems more logical to protect http_parse_host() from being called if the UF_HOST field data is not set anyway than to memset the structure. In any case the patches are compatible and can both be applied for a robust fix of the invalid memory access issue.

http_parser.c
```diff
@@ -2376,10 +2376,8 @@ http_parser_parse_url(const char *buf, size_t buflen, int is_connect,
 
   /* host must be present if there is a schema */
   /* parsing http:///toto will fail */
-  if ((u->field_set & ((1 << UF_SCHEMA) | (1 << UF_HOST))) != 0) {
-    if (http_parse_host(buf, u, found_at) != 0) {
-      return 1;
-    }
+  if ((u->field_set & (1 << UF_SCHEMA)) && (u->field_set & (1 << UF_HOST)) == 0) {
```
indutny: 80 column limit, please move stuff to the new lines.

indutny:
Looks good, but what about failing test? ;)

OnixGH:
@indutny oh sorry, I thought I was being smart by moving the host and port checks together, but the http_parse_host also parses the port and needed to go back up. Test works for me now, and I also addressed the 80 char limit.

http_parser.c
```diff
@@ -2384,7 +2389,7 @@ http_parser_parse_url(const char *buf, size_t buflen, int is_connect,
 
   /* CONNECT requests can only contain "hostname:port" */
   if (is_connect && u->field_set != ((1 << UF_HOST)|(1 << UF_PORT))) {
-    return 1;
+	  return 1;
```
indutny: Unnecessary change? ;)

indutny:
Looks awesome! May I ask you to submit some kind of test for it?

OnixGH:
I checked that the test suite reports a failure with the old code and OK with the new code.

indutny:
Aaah, I see :) Thanks.

indutny:
Landed in f6f436a12edf209be806be8833bced9419d985ca, thank you!

OnixGH:
:+1:

PR #247, opened 2015-05-19
Fix dynamic library extension for OS X

guoxiao:
The extension of dynamic library in OS X should be `.dylib` rather than `.so`.

Here is some explainations about this: http://www.finkproject.org/doc/porting/porting.en.html#shared.lib-and-mod

> Mach-O shared libraries have the file type MH_DYLIB and carry the extension .dylib. They can be linked against with the usual static linker flags, e.g. -lfoo for libfoo.dylib. However, they can not be loaded as a module. (Side note: Shared libraries can be loaded dynamically through an API. However, that API is different from the API for bundles and the semantics make it useless for an dlopen() emulation. Most notably, shared libraries can not be unloaded.)

> Loadable modules are called "bundles" in Mach-O speak. They have the file type MH_BUNDLE. Since no component involved cares about it, they can carry any extension. The extension .bundle is recommended by Apple, but most ported software uses .so for the sake of compatibility. Bundles can be dynamically loaded and unloaded via dyld APIs, and there is a wrapper that emulates dlopen() on top of that API. It is not possible to link against bundles as if they were shared libraries. However, it is possible that a bundle is linked against real shared libraries; those will be loaded automatically when the bundle is loaded.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,8 +19,13 @@
 # IN THE SOFTWARE.
 
 PLATFORM ?= $(shell sh -c 'uname -s | tr "[A-Z]" "[a-z]"')
+ifeq (darwin,$(PLATFORM))
+SONAME ?= libhttp_parser.2.5.0.dylib
+SOEXT ?= dylib
+else
 SONAME ?= libhttp_parser.so.2.5.0
-
+SOEXT ?= so
+endif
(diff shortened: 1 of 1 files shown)


indutny:
Landed in 2896229, thank you!

PR #252, opened 2015-06-12
use ARRAY_SIZE instead of manual array size calculation

deleisha:
Reuse the already defined ARRAY_SIZE macro instead of manual array size calculation


indutny:
LGTM, sorry for delay!

indutny:
Landed in ab0b162

indutny:
Thank you so much!

PR #253, opened 2015-06-18
Support IPv6 Zone ID as per RFC 6874

tatsuhiro-t:
IPv6 scoped address contains Zone ID, which is appended to IPv6
address after '%' separator.  RFC 6874 says that Zone ID after '%'
must consist of 1*(unreserved or pct-encoded).  This commit adds this
IPv6 Zone ID support.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -400,6 +400,8 @@ enum http_host_state
   , s_http_host
   , s_http_host_v6
   , s_http_host_v6_end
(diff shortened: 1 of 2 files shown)


tatsuhiro-t:
Updated patch, to rule out the case where there is just zone ID.

indutny:
Thank you so much! Landed in 7d75dd7, sorry for delay.

PR #254, opened 2015-06-18
Tolerate non-compliant status line responses

daeon:
Tolerate web servers which do not return a status message in the return response.

I have noticed this usse on several websites such downloads from mediafire.com

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -881,10 +881,9 @@ reexecute:
               UPDATE_STATE(s_res_status_start);
               break;
             case CR:
-              UPDATE_STATE(s_res_line_almost_done);
-              break;
             case LF:
-              UPDATE_STATE(s_header_field_start);
+              UPDATE_STATE(s_res_status_start);
+              REEXECUTE();
               break;
             default:
               SET_ERRNO(HPE_INVALID_STATUS);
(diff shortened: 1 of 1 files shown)


indutny:
Do you have a test case?

jasnell:
@daeon ... is this something you still wish to pursue? If so, the PR needs a test case as @indutny requested.

fake666:
as tomcat 8.5 dropped the reason-phrase, i would like to help getting this in - i'll try to write a unit test. the on_status callback is not called at the moment when there is an empty reason phrase...

there is no definite statement that the reason phrase has to be present, here's a discussion: 
https://github.com/spring-projects/spring-boot/issues/6548

tomcat bug:
https://bz.apache.org/bugzilla/show_bug.cgi?id=60362 - fixed, but not released yet

bnoordhuis:
@fake666 Go for it.  Can you add some regression tests to test.c?  Comment if you have questions on how to best approach that and I'll walk you through it.

Apropos this:

> the on_status callback is not called at the moment when there is an empty reason phrase...

Does your comment apply to the current tip-of-tree or this pull request?

fake666:
> Does your comment apply to the current tip-of-tree or this pull request?

sorry: it applies to the current tip-of-tree, this pull request fixes it ;)

as i can't append the unit test to this pull request, i'll open a new one, referencing (and, of course, crediting) this one.

bnoordhuis:
Closing, addressed in 3b0da34 (ref #367.)

PR #260, opened 2015-07-16
Support ACL (WebDAV, RFC3744, Section 8.1)

Hywan:
This PR adds support for the following HTTP method:

  * `ACL` http://tools.ietf.org/html/rfc3744#section-8.1.

This is part of the WebDAV protocol.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -959,6 +959,7 @@ reexecute:
         parser->method = (enum http_method) 0;
         parser->index = 1;
         switch (ch) {
+          case 'A': parser->method = HTTP_ACL; break;
           case 'B': parser->method = HTTP_BIND; break;
           case 'C': parser->method = HTTP_CONNECT; /* or COPY, CHECKOUT */ break;
           case 'D': parser->method = HTTP_DELETE; break;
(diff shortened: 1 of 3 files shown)


indutny:
Is there any other WebDAV methods that you'd like to add? ;) I'd rather do a single commit

Hywan:
@indutny Next methods (related to WebDAV or other RFC) are present in another commits and need a little refactoring, so I would be glad if you could merge this one first, would be a lot more easier here :-].

indutny:
@Hywan I was thinking about it for a couple of days now and have a suggestion.

What do you think about moving parts of the method parsing state machine into separate functions? I really dislike how it blows up the main parser body.

Hywan:
@indutny Could be definitively more testable, for sure. Moreover, I suspect this function to be wrong but magically works for all our usecases :-p.

Hywan:
@indutny Do we have “speed-test”?

indutny:
@Hywan kind of, `make bench`

Hywan:
@indutny Ok… So it's up to you to take the decision ;-). I agree to try a refactoring on my spare time.

indutny:
I'm all for it :)

indutny:
Thank you!

Hywan:
What about this PR though?

indutny:
@Hywan maybe we could do it as a part of refactor?

Hywan:
Actually I need it for a current project, so I would be glad if it could be merged right now and then I will include my next PR in the refactoring (as separated commits).

indutny:
Here you go: eb5e992, thank you!

Hywan:
Thank you so muc!

PR #261, opened 2015-07-22
fixed compile error C2143 for vs2012

sebastian-philipp:
This is just a compile fix


indutny:
Landed in e01811e, thank you!

PR #266, opened 2015-09-27
methods: add link and unlink

jasnell:
LINK and UNLINK are registered methods and are in real world use.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -1000,6 +1000,13 @@ reexecute:
           UPDATE_STATE(s_req_spaces_before_url);
         } else if (ch == matcher[parser->index]) {
           ; /* nada */
+        } else if (parser->method == HTTP_LOCK) {
+          if (parser->index == 1 && ch == 'I') {
+            parser->method = HTTP_LINK;
+          } else {
+            SET_ERRNO(HPE_INVALID_METHOD);
+            goto error;
+          }
         } else if (parser->method == HTTP_CONNECT) {
           if (parser->index == 1 && ch == 'H') {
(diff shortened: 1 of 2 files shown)


http_parser.c
```diff
@@ -123,7 +123,7 @@ do {                                                                 \
     FOR##_mark = NULL;                                               \
   }                                                                  \
 } while (0)
-
```
indutny: Unnecessary style change?

http_parser.h
```diff
@@ -149,7 +152,7 @@ enum flags
 
 
 /* Map for errno-related constants
- *
```
indutny: Ditto.

indutny:
@jasnell are these WebDAV methods? Where are they specified.

http_parser.c
jasnell: Grr.. Thanks. Missed those. Looks like I need to tweak my atom editor
settings so it's not doing these silly style edits

jasnell:
Originally, in RFC2068. They were pulled from 2616 because of lack of use.
I introduced a new I-D to update their definition and RFC 7237 added them
to the iana registry. I had let my ID expire but recently found some real
world implementation that led me to update to ID.

jasnell:
@indutny ... ok, unnecessary style changes removed.

indutny:
Test?

jasnell:
I'll get one added!

dolmen:
I just submitted #267 that has tests.

jasnell:
@dolmen ... thank you!

jasnell:
Closed in favor of #267

PR #279, opened 2016-02-10
Bump to 2.6.1

jbergstroem:
This merges the bump from nodejs and fixes an issue with compiling tests.

/cc=@jasnell

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -435,6 +435,12 @@ enum http_host_state
   (IS_ALPHANUM(c) || (c) == '.' || (c) == '-' || (c) == '_')
 #endif
 
+/**
+ * Verify that a char is a valid visible (printable) US-ASCII
+ * character or %x80-FF
+ **/
+#define IS_HEADER_CHAR(ch)                                                     \
+  (ch == CR || ch == LF || ch == 9 || (ch > 31 && ch != 127))
 
 #define start_state (parser->type == HTTP_REQUEST ? s_start_req : s_start_res)
 
@@ -639,7 +645,8 @@ size_t http_parser_execute (http_parser *parser,
   const char *body_mark = 0;
   const char *status_mark = 0;
   enum state p_state = (enum state) parser->state;
-
+  const unsigned int lenient = parser->lenient_http_headers;
+  
   /* We're in an error state. Don't bother doing anything. */
   if (HTTP_PARSER_ERRNO(parser) != HPE_OK) {
     return 0;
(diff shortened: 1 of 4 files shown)


jasnell:
LGTM

jbergstroem:
fyi green: https://travis-ci.org/nodejs/http-parser/builds/108174614

indutny:
LGTM

jbergstroem:
Feel free to merge if you think its ready!

jasnell:
Landed!

http_parser.c
```diff
@@ -435,6 +435,12 @@ enum http_host_state
   (IS_ALPHANUM(c) || (c) == '.' || (c) == '-' || (c) == '_')
 #endif
 
+/**
+ * Verify that a char is a valid visible (printable) US-ASCII
+ * character or %x80-FF
+ **/
+#define IS_HEADER_CHAR(ch)                                                     \
+  (ch == CR || ch == LF || ch == 9 || (ch > 31 && ch != 127))
```
craig65535: Does this comparison work for values >= 128? `ch` is a signed char so those values would be interpreted as -128..-1 and not valid.

Also, in https://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html#sec2.2 :

```
       CHAR           = <any US-ASCII character (octets 0 - 127)>
[...]
       CTL            = <any US-ASCII control character
                        (octets 0 - 31) and DEL (127)>
[...]
       SP             = <US-ASCII SP, space (32)>
       HT             = <US-ASCII HT, horizontal-tab (9)>
[...]
HTTP/1.1 defines the sequence CR LF as the end-of-line marker for all protocol elements except the entity-body (see appendix 19.3 for tolerant applications). The end-of-line marker within an entity-body is defined by its associated media type, as described in section 3.7.

       CRLF           = CR LF

HTTP/1.1 header field values can be folded onto multiple lines if the continuation line begins with a space or horizontal tab. All linear white space, including folding, has the same semantics as SP. A recipient MAY replace any linear white space with a single SP before interpreting the field value or forwarding the message downstream.

       LWS            = [CRLF] 1*( SP | HT )

The TEXT rule is only used for descriptive field contents and values that are not intended to be interpreted by the message parser. Words of *TEXT MAY contain characters from character sets other than ISO- 8859-1 [22] only when encoded according to the rules of RFC 2047 [14].

       TEXT           = <any OCTET except CTLs,
                        but including LWS>

A CRLF is allowed in the definition of TEXT only as part of a header field continuation. It is expected that the folding LWS will be replaced with a single SP before interpretation of the TEXT value. 
```

My reading is that if a CR or LF char is present, it must be in a pair (CRLF) and then must be followed by either a SP or a HT.

But overall this does over the set of valid characters, so maybe that doesn't matter. Although I would consider adding a definition for HT instead of 9.

craig65535: Looks like this was addressed in #283, which is good, but... does nobody read these pull request comments? Should I have made my own PR instead?

jasnell:
@craig65535 ... the comment had been buried in my github notifications. Actually just saw it a few minutes after opening #283.

PR #283, opened 2016-02-15
Fix regression in IS_HEADER_CHAR check

jasnell:
Should have been checking against unsigned char as opposed to char.
bump to v2.6.2
/cc @indutny

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -440,7 +440,7 @@ enum http_host_state
  * character or %x80-FF
  **/
 #define IS_HEADER_CHAR(ch)                                                     \
-  (ch == CR || ch == LF || ch == 9 || (ch > 31 && ch != 127))
+  (ch == CR || ch == LF || ch == 9 || ((unsigned char)ch > 31 && ch != 127))
 
 #define start_state (parser->type == HTTP_REQUEST ? s_start_req : s_start_res)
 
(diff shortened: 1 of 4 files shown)


http_parser.c
```diff
@@ -440,7 +440,7 @@ enum http_host_state
  * character or %x80-FF
  **/
 #define IS_HEADER_CHAR(ch)                                                     \
-  (ch == CR || ch == LF || ch == 9 || (ch > 31 && ch != 127))
+  (ch == CR || ch == LF || ch == 9 || ((unsigned char)ch > 31 && ch != 127))
```
indutny: Alright, nit: `(unsigned char) ch`.

indutny:
Needs test. Otherwise LGTM

silentroach:
strange workflow, you just fixed it directly in node with more code and test.

why not just sync that changes here?

jasnell:
We have a release for 4.3.1 coming out later on today and I wanted to make sure the regression fix was included in that. Otherwise it would have landed here first. Unfortunately I'm tied up at a conference all day so I won't be able to get the additional tests for this implemented until tomorrow (unless someone wants to beat me to it and open a PR against my dev branch ;-) ... hint hint nudge nudge). Then we can get this updated here and get the downstream and upstream sync'd.

jasnell:
@indutny @jbergstroem ... test case added. I'll squash the commits down before landing. PTAL

jasnell:
Closed in favor of #287

PR #285, opened 2016-02-20
Introduce raw data callbacks

RandoMan70:
These callbacks can be used when developing transparent HTTP traffic analysis and filtering. 
The main goal of it is to handle transmitted data as-is, but having ability to distinguish data between headers, bodies and different requests.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -56,11 +56,30 @@ do {                                                                 \
   parser->http_errno = (e);                                          \
 } while(0)
 
+#define CALLBACK_RAW_ON_RETURN(V)                                    \
+ do {                                                                \
+   if (HTTP_PARSER_ERRNO(parser) == HPE_OK) {                        \
+     if ((p >= raw_mark) && (raw_mark < data + len) && ((ptrdiff_t)(V) == (ptrdiff_t)(len))) { \
+ 	  if (p_state <= s_header_almost_done) {                         \
+ 	    if (settings->on_header_raw) {                               \
+ 		  settings->on_header_raw(parser, raw_mark, len - (raw_mark - data));   \
+ 	    }                                                            \
+ 	  } else {                                                       \
+ 	    if (settings->on_body_raw) {                                 \
+ 		  settings->on_body_raw(parser, raw_mark, len - (raw_mark - data));   \
+   	    }                                                            \
+   	  }                                                              \
+     }                                                               \
+   }                                                                 \
+ }                                                                   \
+ while (0);
+
 #define CURRENT_STATE() p_state
 #define UPDATE_STATE(V) p_state = (enum state) (V);
 #define RETURN(V)                                                    \
 do {                                                                 \
   parser->state = CURRENT_STATE();                                   \
+  CALLBACK_RAW_ON_RETURN(V);                                         \
   return (V);                                                        \
 } while (0);
 #define REEXECUTE()                                                  \
@@ -80,6 +99,17 @@ do {                                                                 \
 #define CALLBACK_NOTIFY_(FOR, ER)                                    \
 do {                                                                 \
   assert(HTTP_PARSER_ERRNO(parser) == HPE_OK);                       \
+                                                                     \
+  if (e_on_##FOR == e_on_message_complete) {                         \
+     if ((HTTP_PARSER_ERRNO(parser) == HPE_OK) &&                    \
+         (settings->on_body_raw) &&                                  \
+         (p >= raw_mark) &&                                          \
+         (raw_mark < data + len))                                    \
+   {                                                                 \
+       settings->on_body_raw(parser, raw_mark, p - raw_mark + 1);    \
+       raw_mark = p + 1;                                             \
+     }                                                               \
+   }                                                                 \
                                                                      \
   if (LIKELY(settings->on_##FOR)) {                                  \
     parser->state = CURRENT_STATE();                                 \
@@ -633,6 +663,7 @@ size_t http_parser_execute (http_parser *parser,
   char c, ch;
   int8_t unhex_val;
   const char *p = data;
+  const char *raw_mark = data;
   const char *header_field_mark = 0;
   const char *header_value_mark = 0;
   const char *url_mark = 0;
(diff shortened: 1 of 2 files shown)


indutny:
Thank you for submitting this! May I ask you to write a short example in C to demonstrate how this API could be used?

RandoMan70:
I just figured out that we need to change position of callbacks in http_parser_settings structure to keep it backward-compatible. I'll update PR and then will do short example.

RandoMan70:
Short example on how it can be used:

```
std::string hdr;

static int headers_raw(http_parser*parser, const char *at, size_t length)
{
	hdr.append(at, length)
}

static int headers_complete(http_parser*parser)
{
	do_anything_when_headers_received();
	send(fd, hdr.c_str(), hdr.size());
	hdr.clear();
}

static int body_raw(http_parser*parser, const char *at, size_t length)
{
	send(fd, at, length);
}
```
But when developing complex processing, we need keep in mind, that `on_header_raw()` callback can be received **before** `on_message_begin`. It's easy to reproduce - just send alone newline symbol  before sending request. It happening because of non-caching nature of HTTP parser.

bnoordhuis:
The numerous style errors alone make this pull request unsuitable for inclusion in its current shape.  As to the functionality, I have no strong opinion on whether it's a good addition.

Something to keep in mind is that http_parser is more strict in what it accepts than most browsers.  If you're going to use it for MitM-style analysis or filtering, you may have a hard time.

RandoMan70:
@bnoordhuis, I don't  insist, but these additions extend the limits of applicability of the parser. 
Writing transparent proxy with support of keep-alive and pipelining very difficult (if possible at all) without raw data callbacks. Using raw callbacks it becomes quite easy.

bnoordhuis:
I'll let @indutny make the call on whether or not to accept this pull request.

> Writing transparent proxy with support of keep-alive and pipelining very difficult (if possible at all) without raw data callbacks.

Is that really true?  It's my experience that transparent proxies still frequently need to rewrite parts of the request and the response.

RandoMan70:
@bnoordhuis, there are different tasks for proxying. Main goal of raw callbacks is to keep control of request/response sequence without need of reassembling stream before forwarding.

http_parser.c
```diff
@@ -81,6 +107,17 @@ do {                                                                 \
 do {                                                                 \
   assert(HTTP_PARSER_ERRNO(parser) == HPE_OK);                       \
                                                                      \
+  if (e_on_##FOR == e_on_message_complete) {                         \
+     if ((HTTP_PARSER_ERRNO(parser) == HPE_OK) &&                    \
+         (settings->on_body_raw) &&                                  \
+         (p >= raw_mark) &&                                          \
+         (raw_mark < data + len))                                    \
+   {                                                                 \
```
indutny: Please place braces on the same line as the main expression.

http_parser.c
```diff
@@ -81,6 +107,17 @@ do {                                                                 \
 do {                                                                 \
   assert(HTTP_PARSER_ERRNO(parser) == HPE_OK);                       \
                                                                      \
+  if (e_on_##FOR == e_on_message_complete) {                         \
+     if ((HTTP_PARSER_ERRNO(parser) == HPE_OK) &&                    \
+         (settings->on_body_raw) &&                                  \
+         (p >= raw_mark) &&                                          \
```
indutny: No need in extra parens for every subcondition.

http_parser.c
```diff
@@ -1847,6 +1885,18 @@ size_t http_parser_execute (http_parser *parser,
...
            parser->method == HTTP_CONNECT);
 
+        /* Here we call final headers_raw. It's based on different variables,
+         * so we can't use CALLBACK_DATA.
+         */
+        if ((HTTP_PARSER_ERRNO(parser) == HPE_OK) && 
+            (settings->on_header_raw) && 
+            (p >= raw_mark) && 
+            (raw_mark < data + len)) 
+        {
```
indutny: Please move it to previous line.

http_parser.c
```diff
@@ -56,11 +56,37 @@ do {                                                                 \
   parser->http_errno = (e);                                          \
 } while(0)
 
+#define CALLBACK_RAW_ON_RETURN(V)                                    \
+ do {                                                                \
```
indutny: Odd indent, please put it at the same ident level as above defines.

http_parser.c
```diff
@@ -56,11 +56,37 @@ do {                                                                 \
...
+         ((ptrdiff_t)(V) == (ptrdiff_t)(len)))                       \
+     {                                                               \
+       if (p_state <= s_header_almost_done) {                        \
+         if (settings->on_header_raw) {                              \
+           settings->on_header_raw(parser,                           \
+                                   raw_mark,                         \
+                                   len - (raw_mark - data));         \
+         }                                                           \
+       } else {                                                      \
+         if (settings->on_body_raw) {                                \
```
indutny: Could be merged with previous line into `else if`

indutny:
@RandoMan70 I don't really object to this change, but on other hand try to avoid touching that code. It is very sensitive code path, which may ruin the performance of the method because of high register pressure.

What do you think about doing this under some build-time define? Clearly this feature is not something that the most of the users will use, so it will make more sense to do it this way.

socketpair:
Just store that here: https://github.com/OISF/libhtp/pull/123 exactly the same, but in libhtp!

socketpair:
@RandoMan70 can you write me a message ? socketpair@gmail.com

RandoMan70:
@indutny it's possible, sure. But don't sure that it can significantly affect performance - checks executed mosly once per call and per request/response switch.

bnoordhuis:
Not in v2.x since it changes the ABI.  Maybe in 3.0 if and when that is released.

(It could be put behind a build-time flag but that's mighty inconvenient for some downstream users.  Distros already have a hard enough time with the strict and non-strict builds.)

HAT-DK:
Any update on adding this pull request, in a release?

bnoordhuis:
There are no concrete plans for a v3.x release so far.

indutny:
@HAT-DK It isn't mentioned in the readme, but nodejs is moving away from http-parser to [llhttp][0]. If you're interested, I could guide you through porting this PR to it.

[0]: https://github.com/nodejs/llhttp

socketpair:
@asvetlov

Seems, aiohttp should move to this parser too :)

PR #287, opened 2016-02-27
Version bump: http-parser-2.6.2

jbergstroem:
This is an adaption of the bump that was introduced in the nodejs tree. Hopefully we can improve this workflow moving forward since we're adding subtle discrepancies between the codebases.

Also, add a test for obstext characters.

Guessing this could be dropped in favour of #283 (and moving the test commit there). Your call.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,14 +22,14 @@ PLATFORM ?= $(shell sh -c 'uname -s | tr "[A-Z]" "[a-z]"')
 HELPER ?=
 BINEXT ?=
 ifeq (darwin,$(PLATFORM))
-SONAME ?= libhttp_parser.2.6.1.dylib
+SONAME ?= libhttp_parser.2.6.2.dylib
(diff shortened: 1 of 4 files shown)


jbergstroem:
ping @indutny, @jasnell! Again; happy to see #283 merge instead. Your call -- I mostly want a release and hopefully get commits in here moving forward.

jbergstroem:
Just a note; we can't use master to cut 2.6.2 any longer -- there's already been a commit landed that would introduce further discrepancies between the nodejs version of http-parser and this repo. An alternative would be branching off prior the last commit to master or just jump to 2.6.3 and import that to nodejs (making sure all changes follow). I can help out having a look at/review other open pr's if we're looking to get some changes in.

jasnell:
oh... this one works too. LGTM

jasnell:
Landed in dbcda19 and 5651aea

PR #296, opened 2016-03-29
Fix: suppress warning C4244 on MSVC

bc-lee:
MSVC 2015 gives some warnings on http_parser.c with 

> warning C4244: '=' : conversion from 'int' to 'uint16_t', possible loss of data
> It will supress this warning.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -29,6 +29,11 @@
 #include <string.h>
 #include <limits.h>
 
+#ifdef _MSC_VER
+#pragma warning (push)
+#pragma warning (disable:4244) // warning C4244: '=' : conversion from 'int' to 'uint16_t', possible loss of data
+#endif //_MSC_VER
+
 #ifndef ULLONG_MAX
 # define ULLONG_MAX ((uint64_t) -1) /* 2^64-1 */
 #endif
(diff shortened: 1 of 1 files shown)


indutny:
@bc-lee which line produces this warning?

bc-lee:
@indutny In line 2273, 2280, 2292, 2301 (conversion from 'int' to 'uint16_t')
> u->field_data[UF_HOST].off = p - buf;

and, in line 1415, 1520, 1689 (conversion from '__int64' to 'uint32_t') which occurs only if using 64bit MSVC.
> COUNT_HEADER_SIZE(p - start);

indutny:
Thank you for prompt reply! May I ask you if you have any particular reason to fix these warnings? Or if this PR is intended to remove them just for the sake of clean compiler output?

cc @mscdex @jasnell

craig65535:
IMO fixing the code is preferable to suppressing the warning. It's possible that all you need is to cast to uint16_t.

jasnell:
I'm fine with the change, I do agree that we may need to refactor this bit of the code for a bit to avoid the issue in the first place.

misery:
Ping! Any progress here? :-)

See: https://ci.appveyor.com/project/GovernikusAusweisApp2/ausweisapp2/build/1.14.0-6/job/q0a43ytp6b1jfkc4/messages

PR #299, opened 2016-04-16
parser: returning 2 from on_headers_complete

indutny:
Returning `2` from on_headers_complete will tell parser that it
should not expect neither a body nor any futher responses on
this connection. This is useful for handling responses to a
CONNECT request which may not contain `Upgrade` or
`Connection: upgrade` headers.

See: https://github.com/nodejs/node/pull/6198

cc @bnoordhuis @jasnell @mscdex 

See also: https://github.com/nodejs/node/pull/6198#issuecomment-210857637

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -1812,6 +1812,9 @@ reexecute:
             case 0:
               break;
 
+            case 2:
+              parser->upgrade = 1;
+
             case 1:
               parser->flags |= F_SKIPBODY;
               break;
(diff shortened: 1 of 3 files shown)


mscdex:
LGTM if tests pass.

indutny:
@mscdex thank you, this is semver minor, right?

mscdex:
Assuming the original intention was for `CONNECT` to be handled in this way, then I would say yes.

indutny:
Yeah, also existing code will work just as it was before.

indutny:
A question for @nodejs/lts team, do you think this could be backported to v4.x.x? It seems that responses to CONNECT requests are currently a bit broken in LTS

indutny:
The problem is that this is a semver-minor change in http-parser.

indutny:
Alright, enough time to collect feedback. Landing.

indutny:
Landed in 04d28a7, thank you everyone!

PR #300, opened 2016-04-22
Fix error: variable ‘nread’ set but not used

guoxiao:
```
cc  -I. -DHTTP_PARSER_STRICT=0  -Wall -Wextra -Werror -O3  -c test.c -o test.o
test.c: In function ‘test_message_connect’:
test.c:3853:10: error: variable ‘nread’ set but not used [-Werror=unused-but-set-variable]
   size_t nread;
          ^~~~~
cc1: all warnings being treated as errors
```

### DIFF
--- a/test.c
+++ b/test.c
@@ -3850,11 +3850,10 @@ test_message_connect (const struct message *msg)
 {
   char *buf = (char*) msg->raw;
   size_t buflen = strlen(msg->raw);
-  size_t nread;
 
   parser_init(msg->type);
 
-  nread = parse_connect(buf, buflen);
+  parse_connect(buf, buflen);
 
   if (num_messages != 1) {
     printf("\n*** num_messages != 1 after testing '%s' ***\n\n", msg->name);


rtvd:
I see the same. Also, the code looks strange:

`  size_t nread;

  parser_init(msg->type);

  nread = parse_connect(buf, buflen);

  if (num_messages != 1) {
    printf("\n*** num_messages != 1 after testing '%s' ***\n\n", msg->name);
    abort();
  }`

nread is set to the result of "parse_connect" but then "num_messages" is queried. It is counter-intuitive. Also, num_messages is a global variable, which does not help to figure out the logic behind this.

indutny:
@rtvd I agree that the logic in tests could be quite complex to follow. Not sure if I'm willing to take on this and rewrite them at this moment, though.

ploxiln:
This little patch should probably be merged whether or not tests are refactored, right? Because it's causing CI to report test failure?

indutny:
Landed in b2cc8e49f39277b197c6f3592708309ee78c015e, thank you!

indutny:
Sorry for delay.

PR #303, opened 2016-05-03
Allow keep-alive for HTTP/1.1 and later

geek:
Without this fix, HTTP/2.0 wouldn't support keep-alive, but 1.9 would, as would 2.1.


indutny:
@geek the problem is that HTTP/2.0 is using different protocol :) Technically, there is not going to be any new compatible version that will violate this check.

Thanks for contribution, though!

geek:
@indutny thanks for the quick response.

PR #311, opened 2016-05-28
Allow non-standard HTTP method

tatsuhiro-t:
This change adds non-standard HTTP method support.  We added new
callback on_method of type http_data_cb, which notifies incoming
method.  For non-standard HTTP method, method field in struct
http_parser becomes HTTP_METHOD_UNKNOWN.

The non-standard HTTP method is disabled by default, and must be
explicitly enabled by giving -DHTTP_PARSER_METHOD_CB=1 on build time.
This is for backward compatibility; because non-standard HTTP method
is now accepted rather than rejected.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -280,6 +280,7 @@ enum state
   { s_dead = 1 /* important that this is > 0 */
 
   , s_start_req_or_res
+  , s_res_or_resp_mark_H
   , s_res_or_resp_H
   , s_start_res
   , s_res_H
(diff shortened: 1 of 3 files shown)


tatsuhiro-t:
Not sure you like this PR, it is a bit ugly with lots of if-def dance.

The only concern at the moment is that if HTTP_BOTH is used and this non-standard HTTP method support is enabled, on_method will be called even if it later turns out that it is response.

vinniefalco:
Is there a use-case for `HTTP_BOTH`? I can't imagine any real-world scenario where you might get either a request or a response on a connection. As far as I am concerned, this feature could be removed. I would love to hear a counter-example.

nazar-pc:
Can this PR be moved forward somehow? It is quite old, would be nice to have it merged.

PR #320, opened 2016-06-16
Use nullptr if compiler uses c++11

misery:
This will allow "-Wzero-as-null-pointer-constant" in client code.

See: http://www.stroustrup.com/C++11FAQ.html#11

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -29,6 +29,12 @@
 #include <string.h>
 #include <limits.h>
 
+#if __cplusplus > 199711L
+# define NULLPTR nullptr
+#else
+# define NULLPTR 0
+#endif
+
 #ifndef ULLONG_MAX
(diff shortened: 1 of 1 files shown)


indutny:
Hello!

Thank you for submitting this! I believe http_parser is built with a C compiler, not C++. Thus I can't see how it could get in.

Cheers,
Fedor.

misery:
If someone (like us) use http_parser as an embedded resource it can be compiled by c++ compiler.

indutny:
@misery I totally understand. However even the source files have `.c` extension... thus, having C++ primitives in them is probably something that no one would expect to find.

Really sorry to close this, but I don't think that this additional will benefit the most of the users.

cc @mscdex @jasnell what do you think?

mscdex:
I agree with @indutny, however it would be nice if we at least used `NULL` consistently.

PR #323, opened 2016-06-30
simplify parsing M-SEARCH method, group P methods

ploxiln:
This just makes the method verb parsing code a bit more elegant:
- can use same switch-lookup for `-` char case for `M-SEARCH`
- move PROPFIND and PURGE to be next to the other P methods

also, change IS_ALPHA(ch) to just A <= ch <= Z
(very slight optimization, only uppercase will match in switch)

I tested with `./bench` on OS X and in a Linux VM, and performance is pretty much the same with this change, just an insignificant smidge faster. I tested branches in back/forth/back/forth pattern.

```
OS X,     master     :  830174.250000 req/sec,  829930.875000 req/sec
OS X,     this branch:  841326.375000 req/sec,  836593.812500 req/sec

Linux VM, master     : 1164776.500000 req/sec, 1165652.750000 req/sec
Linux VM, this branch: 1179126.250000 req/sec, 1175704.875000 req/sec
```

But this isn't about the performance, just mostly about getting rid of the extra if statement just for the `M-Search` method.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -1007,7 +1007,7 @@ reexecute:
           UPDATE_STATE(s_req_spaces_before_url);
         } else if (ch == matcher[parser->index]) {
           ; /* nada */
-        } else if (IS_ALPHA(ch)) {
+        } else if ((ch >= 'A' && ch <= 'Z') || ch == '-') {
 
           switch (parser->method << 16 | parser->index << 8 | ch) {
 #define XX(meth, pos, ch, new_meth) \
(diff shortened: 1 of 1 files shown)


http_parser.c
```diff
@@ -1007,7 +1007,7 @@ size_t http_parser_execute (http_parser *parser,
           UPDATE_STATE(s_req_spaces_before_url);
         } else if (ch == matcher[parser->index]) {
           ; /* nada */
-        } else if (IS_ALPHA(ch)) {
+        } else if ((ch >= 'A' && ch <= 'Z') || ch == '-') {
```
indutny: Doesn't `IS_ALPHA` also cast the character to the lower case?

ploxiln: It does, just for the check. It does not modify `ch` for later code.

But if the letter is lowercase instead of uppercase, it will (when mixed with the other variables for the fancy select value) never match any case statement. It will always hit the "default" case. Which is the same as not matching this if condition.

indutny: Good point. Thank you!

indutny:
LGTM.

@jasnell or @mscdex : do you guys want to sign off?

ploxiln:
I am willing to make adjustments if desired, fwiw.

indutny:
@ploxiln there is no need in this! I'll land it in 24 hours if no one will reply.

ploxiln:
friendly ping 😁

jasnell:
O.o just saw this! sorry about that... LGTM if @indutny is happy with it.

indutny:
Landed in 1b79aba, thank you! Sorry for very very very long delay 😭

Issue #324, opened 2016-07-03
Content-Length Header


saschaarthur:
Hello,

I just run into in issue with your http_parser.c.

I digged a bit into your Code, noticing that youre using something similiar to a state-machine to match a substring of the server returning Headers. (https://github.com/nodejs/http-parser/blob/master/http_parser.c#L1301)

This approach is fast and works in theorie if youre parsing the rfc7230. But well yeah, the outside world - also in my case not always stick fully to this RFC.

Were using a custom REST implementation and adding additional Headers.

We added for example for furher information "Content-Length-Uncompressed", because we gzipping our content, but need also the real size.

A Change in this commit (https://github.com/nodejs/http-parser/commit/e2e467b91262246b339fb3d80c8408d498b4a43b) (https://github.com/nodejs/http-parser/blame/master/http_parser.c#L1369) broke nodejs for working with our rest in the latest version. 

We now changed this Header Name on our side, but you can and should consider to parse the full header String instead only of a Substring which is a bit dangerous and brings unwanted side-effects.

Fishrock123:
cc @nodejs/http

bnoordhuis:
https://github.com/nodejs/http-parser/pull/325

jasnell:
Yes, this is an unfortunate side effect of the current parsing model. I see @bnoordhuis already has a pr for this specific issue but there are more systemic problems with the current parser that I'll be looking at more in depth starting in the coming weeks.

PR #325, opened 2016-07-04
Fix Content-Length header parsing.

bnoordhuis:
Commit e2e467b ("Update http-parser to 2.6.1") enforces that messages
contain no more than one Content-Length header but it considers any
header that starts with "Content-Length" as a duplicate.

Fixes: https://github.com/nodejs/http-parser/issues/324

R=@jasnell


indutny:
LGTM

jasnell:
LGTM

ploxiln:
This seems kinda important, and many seem to agree it's a good implementation ... what's the hold-up?

indutny:
Sorry for delay. Landed in cacb07d, and released 2.7.1.

PR #326, opened 2016-07-05
add settings flags, refactor lenient_http_headers setting

jasnell:
A previous change during a security fix added the ability to set a `lenient_http_headers` flag on the HTTP Parser instance. The way it was done was a bit of a hack at the time in order to avoid an ABI breaking change. This PR adds a new `uint32_t` flags field to `http_parser_settings` in order to provide a more robust and correct way of passing these kinds of settings into a parser instance.

_Note_: This is an ABI change so it would require a semver-major bump

/cc @indutny

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -645,7 +645,8 @@ size_t http_parser_execute (http_parser *parser,
   const char *body_mark = 0;
   const char *status_mark = 0;
   enum state p_state = (enum state) parser->state;
-  const unsigned int lenient = parser->lenient_http_headers;
+  const unsigned int lenient =
+    (settings->flags & SETTINGS_FLAG_LENIENT_HTTP_HEADERS);
 
   /* We're in an error state. Don't bother doing anything. */
   if (HTTP_PARSER_ERRNO(parser) != HPE_OK) {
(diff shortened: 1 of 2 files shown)


http_parser.c
```diff
@@ -645,7 +645,8 @@ size_t http_parser_execute (http_parser *parser,
   const char *body_mark = 0;
   const char *status_mark = 0;
   enum state p_state = (enum state) parser->state;
-  const unsigned int lenient = parser->lenient_http_headers;
+  const unsigned int lenient =
+    (settings->flags & SETTINGS_FLAG_LENIENT_HTTP_HEADERS);
```
indutny: Superfluous parens?

http_parser.h
```diff
@@ -218,6 +218,10 @@ enum http_errno {
 /* Get an http_errno value from an http_parser */
 #define HTTP_PARSER_ERRNO(p)            ((enum http_errno) (p)->http_errno)
 
+enum http_parser_settings_flags {
+  SETTINGS_FLAG_DEFAULT = 0,
```
indutny: `HTTP_SETTINGS_...`?

PR #334, opened 2016-09-14
build: vms and os400 support

jlesk:


http_parser.c
```diff
@@ -406,6 +397,118 @@ enum http_host_state
...
+    [HTTP_UNSUBSCRIBE] = "UNSUBSCRIBE",
+    [HTTP_PATCH] = "PATCH",
+    [HTTP_PURGE] = "PURGE",
+    [HTTP_MKCALENDAR] = "MKCALENDAR",
+    [HTTP_LINK] = "LINK",
+    [HTTP_UNLINK] = "UNLINK"
+};
+
+
+static const char *errno_name[] = {
```
indutny: These symbols are globally available everywhere, where `http_parser.h` is included. I'm not sure it is a good idea to declare it with such unprefixed names.

indutny:
I'm sorry, but from what it looks like - the code just bloated significantly. I doesn't look like we will accept this patch at the time.

cc @nodejs/http just in case.

jlesk:
Thanks for your comment. The nested macro doesn't work on os400.

indutny:
Is there any chance to use a newer compiler on os400?

jlesk:
I think it is already the most recent one.

indutny:
cc @bnoordhuis maybe

http_parser.c
```diff
@@ -461,15 +564,6 @@ do {                                                                 \
 
 
 /* Map errno values to strings for human-readable output */
```
craig65535: You may want to move this comment

http_parser.h
```diff
@@ -161,58 +150,42 @@ enum flags
  *
  * The provided argument should be a macro that takes 2 arguments.
```
craig65535: Seems this comment is no longer valid

bnoordhuis:
Closing.  Conflicts and unlikely to get merged.  Thanks anyway.

FWIW, http_parser is known to work on z/os nowadays.

PR #335, opened 2016-09-23
Remove duplicate code

jingsam:


indutny:
Sorry, but this is not a duplicate code. `%.*s` needs two arguments: number of characters to print, and the pointer to the string itself.

PR #337, opened 2016-10-06
Add HTTP_STATUS_MAP(XX) and enum http_status

npmccallum:
This patch provides an enum for the standardized HTTP status codes.
Additionally, the HTTP_STATUS_MAP(XX) can be used for other purposes as well,
such as code-to-name lookups and code-based switch statements.

### DIFF
--- a/http_parser.h
+++ b/http_parser.h
@@ -90,6 +90,76 @@ typedef int (*http_data_cb) (http_parser*, const char *at, size_t length);
 typedef int (*http_cb) (http_parser*);
 
 
+/* Status Codes */
+#define HTTP_STATUS_MAP(XX)                                                 \
+  XX(100, CONTINUE,                        Continue)                        \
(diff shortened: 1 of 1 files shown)


npmccallum:
For some background, I'm using this in the http://github.com/latchset/tang project. I'm using it for two reasons. First, having named constants for the error codes makes the code more readable. Second, I am using the HTTP_STATUS_MAP(XX) macro in order to create a switch statement for generating reply messages with the appropriate status code text.

I found it useful, and I'm sure that other people will too.

jhrozek:
A general +1 to this PR, we're hardcoding a similar lookup table in SSSD as well.

indutny:
@npmccallum if no one will reply in 2-3 days - please ping me again, I'll make sure to land it immediately!

mscdex:
LGTM

indutny:
Landed in 335850f, thank you!

PR #342, opened 2016-10-19
Various optimizations

shekhei:
Hi, basically I was just messing around and did some benchmarking, found some interesting things, so I made some changes. Mostly to do with removing some of the branches and created a SIMD version of finding CRLF that doesn't need to walk the whole payload twice(worst case)

Below are the benchmark results
#### Clang

| time(master) | req/s(master) | time(new) | req/s(new) | percent |
| --- | --- | --- | --- | --- |
| 5.761632 | 867809.687500 | 5.104409 | 979545.312500 | 11.4069 |
| 5.769115 | 866684.062500 | 5.104001 | 979623.625000 | 11.5289 |
| 5.803461 | 861554.875000 | 5.086198 | 983052.625000 | 12.3592 |
| 5.779534 | 865121.687500 | 5.082990 | 983672.937500 | 12.0519 |
| 5.762670 | 867653.375000 | 5.092542 | 981827.875000 | 11.6288 |
#### GCC 4.9

| time(master) | req/s(master) | time(new) | req/s(new) | percent |
| --- | --- | --- | --- | --- |
| 4.297466 | 1163476.375000 | 3.787979 | 1319965.000000 | 11.8555 |
| 4.159397 | 1202097.250000 | 3.774015 | 1324849.000000 | 9.26533 |
| 4.230322 | 1181943.125000 | 3.788900 | 1319644.250000 | 10.4347 |
| 4.214539 | 1186369.375000 | 3.772407 | 1325413.750000 | 10.4906 |
| 4.162938 | 1201074.750000 | 3.763118 | 1328685.375000 | 9.6042 |
### ubuntu 14.04 ( Intel Core i7-6700 @ 3.40GHz )

| time(master) | req/s(master) | time(new) | req/s(new) | percent |
| --- | --- | --- | --- | --- |
| 3.186016 | 1569358.000000 | 2.815905 | 1775628.000000 | 11.6167 |
| 3.186449 | 1569144.875000 | 2.846761 | 1756382.125000 | 10.6604 |
| 3.190864 | 1566973.625000 | 2.877822 | 1737425.125000 | 9.81057 |
| 3.183230 | 1570731.625000 | 2.813425 | 1777193.250000 | 11.6173 |
| 3.190520 | 1567142.625000 | 2.849443 | 1754728.875000 | 10.6903 |

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -495,6 +495,8 @@ int http_message_needs_eof(const http_parser *parser);
 # if defined(_MSC_VER) || \
     (defined(__SSE2__) && defined(__GNUC__))
 #   define USE_INTRISICS_CRLF 1
+# else
+#   define USE_INTRISICS_CRLF 0
 # endif
 #endif
 
@@ -520,14 +522,14 @@ const char* find_crlf(const char* p, const char* data, size_t len) {
   uint32_t result = 0;
 
   while ((uintptr_t)p & 15 && p <= lastp) {
-    if ( *p == CR || *p == LF ) {
+    if (*p == CR || *p == LF) {
       return p;
     }
     ++p;
   }
-  if ( lastp - p < 32 ) {
+  if (lastp - p < 32) {
     while (p <= lastp) {
-      if ( *p == CR || *p == LF ) {
+      if (*p == CR || *p == LF) {
         return p;
       }
       ++p;
@@ -544,7 +546,7 @@ const char* find_crlf(const char* p, const char* data, size_t len) {
     }
     if (!result) {
       while (p <= lastp) {
-        if ( *p == CR || *p == LF ) {
+        if (*p == CR || *p == LF) {
           return p;
         }
         ++p;
--- a/http_parser.c
+++ b/http_parser.c
@@ -481,14 +481,14 @@ static struct {
 
 int http_message_needs_eof(const http_parser *parser);
 
-const char* findCRLF(const char* p, const char* data, size_t len);
+const char* find_crlf(const char* p, const char* data, size_t len);
 
 
 #if defined(__SSE2__) && defined(__GNUC__) 
 
 #include <emmintrin.h>
 
-const char* findCRLF(const char* p, const char* data, size_t len) {
+const char* find_crlf(const char* p, const char* data, size_t len) {
   const char* lastp = MIN(data+len, HTTP_MAX_HEADER_SIZE+p);
 
   int32_t result = 0;
@@ -539,7 +539,7 @@ const char* findCRLF(const char* p, const char* data, size_t len) {
 
 #else
 
-const char* findCRLF(const char* p, const char* data, size_t len) {
+const char* find_crlf(const char* p, const char* data, size_t len) {
   const char* p_cr;
   const char* p_lf;
   size_t limit = data + len - p;
@@ -1322,7 +1322,7 @@ reexecute:
       case s_header_field:
       {
         const char* start = p;
-HEADER_FIELD_BEGIN:
+header_field_begin:
         if ( parser->header_state == h_general ) {
           for (; p != data + len; p++) {
             ch = *p;
@@ -1342,7 +1342,7 @@ HEADER_FIELD_BEGIN:
             switch (parser->header_state) {
               case h_general:
                 --p;
-                goto HEADER_FIELD_BEGIN;
+                goto header_field_begin;
                 break;
 
               case h_C:
@@ -1570,7 +1570,7 @@ HEADER_FIELD_BEGIN:
           switch (h_state) {
             case h_general:
             {
-              p = findCRLF(p, data, len);
+              p = find_crlf(p, data, len);
               --p;
 
               break;
--- a/http_parser.c
+++ b/http_parser.c
@@ -492,7 +492,11 @@ const char* find_crlf(const char* p, const char* data, size_t len);
 #include <emmintrin.h>
 #endif
 
-#if defined(_MSC_VER) || (defined(__SSE2__) && defined(__GNUC__))
+#ifndef USE_INTRISICS_CRLF
+# define USE_INTRISICS_CRLF defined(_MSC_VER) || (defined(__SSE2__) && defined(__GNUC__))
+#endif
+
+#if USE_INTRISICS_CRLF
 
 int32_t get_crlf_mask(const char* p) {
   /* [ c, 0, 0, 0, 0, 0 .. 0 ] */
@@ -523,7 +527,7 @@ const char* find_crlf(const char* p, const char* data, size_t len) {
   if (!result) {
     while(!result && lastp >= p+32) {
       p += 32;
-	  result = get_crlf_mask(p + 16) << 16 | get_crlf_mask(p);
+      result = get_crlf_mask(p + 16) << 16 | get_crlf_mask(p);
     }
     if (!result) {
       return data + len;
--- a/http_parser.c
+++ b/http_parser.c
@@ -481,6 +481,88 @@ static struct {
 
 int http_message_needs_eof(const http_parser *parser);
 
+const char* findCRLF(const char* p, const char* data, size_t len);
+
+
+#ifdef __SSE2__ 
+
+#include <emmintrin.h>
+
+const char* findCRLF(const char* p, const char* data, size_t len) {
+  const char* lastp = MIN(data+len, HTTP_MAX_HEADER_SIZE+p);
+
+  long result = 0;
+  __m128i v1, v2;
+  __m128i vCR = _mm_set_epi32(0x0a0a0a0a, 0x0a0a0a0a,0x0a0a0a0a, 0x0a0a0a0a); // [ c, 0, 0, 0, 0, 0 .. 0 ]
+
+  __m128i vLF = _mm_set_epi32(0x0d0d0d0d, 0x0d0d0d0d,0x0d0d0d0d, 0x0d0d0d0d); // [ c, 0, 0, 0, 0, 0 .. 0 ]
+
+  size_t alignment = (long)p & 15;
+  p -= alignment;
+
+  v1 = *((__m128i*)(p));
+  v2 = _mm_cmpeq_epi8(vCR, v1);
+  v1 = _mm_cmpeq_epi8(vLF, v1);
+  v2 = _mm_or_si128(v1, v2);
+  result = _mm_movemask_epi8(v2);
+
+  v1 = *((__m128i*)(p + 16));
+  v2 = _mm_cmpeq_epi8(vCR, v1);
+  v1 = _mm_cmpeq_epi8(vLF, v1);
+  v2 = _mm_or_si128(v1, v2);
+  result = (_mm_movemask_epi8(v2) << 16 ) | result;
+  result = (result >> alignment) << alignment;
+
+  if ( !result ) {
+    while( !result && lastp >= (p+32) ) {
+      p += 32;
+      v1 = *((__m128i*)(p));
+      v2 = _mm_cmpeq_epi8(vCR, v1);
+      v1 = _mm_cmpeq_epi8(vLF, v1);
+      v2 = _mm_or_si128(v1, v2);
+      result = _mm_movemask_epi8(v2);
+
+      v1 = *((__m128i*)(p+16));
+      v2 = _mm_cmpeq_epi8(vCR, v1);
+      v1 = _mm_cmpeq_epi8(vLF, v1);
+      v2 = _mm_or_si128(v1, v2);
+      result = (_mm_movemask_epi8(v2) << 16 ) | result;
+    }
+    if ( !result ) { return data+len; }
+  }
+  p += __builtin_ctz(result);
+  if ( p >= lastp ) {
+    return data+len;
+  }
+  return p;
+}
+
+#else
+
+const char* findCRLF(const char* p, const char* data, size_t len) {
+  const char* p_cr;
+  const char* p_lf;
+  size_t limit = data + len - p;
+
+  limit = MIN(limit, HTTP_MAX_HEADER_SIZE);
+
+  p_cr = (const char*) memchr(p, CR, limit);
+  p_lf = (const char*) memchr(p, LF, limit);
+  if (p_cr != NULL) {
+    if (p_lf != NULL && p_cr >= p_lf)
+      p = p_lf;
+    else
+      p = p_cr;
+  } else if (UNLIKELY(p_lf != NULL)) {
+    p = p_lf;
+  } else {
+    p = data + len;
+  }
+  return p;
+}
+
+#endif
+
 /* Our URL parser.
  *
  * This is designed to be shared by http_parser_execute() for URL validation,
@@ -1488,24 +1570,7 @@ HEADER_FIELD_BEGIN:
           switch (h_state) {
             case h_general:
             {
-              const char* p_cr;
-              const char* p_lf;
-              size_t limit = data + len - p;
-
-              limit = MIN(limit, HTTP_MAX_HEADER_SIZE);
-
-              p_cr = (const char*) memchr(p, CR, limit);
-              p_lf = (const char*) memchr(p, LF, limit);
-              if (p_cr != NULL) {
-                if (p_lf != NULL && p_cr >= p_lf)
-                  p = p_lf;
-                else
-                  p = p_cr;
-              } else if (UNLIKELY(p_lf != NULL)) {
-                p = p_lf;
-              } else {
-                p = data + len;
-              }
+              p = findCRLF(p, data, len);
               --p;
 
               break;
--- a/http_parser.c
+++ b/http_parser.c
@@ -22,6 +22,7 @@
  * IN THE SOFTWARE.
  */
 #include "http_parser.h"
+#include "http_parser_internal.h"
 #include <assert.h>
 #include <stddef.h>
 #include <ctype.h>
@@ -483,8 +484,6 @@ static struct {
 
 int http_message_needs_eof(const http_parser *parser);
 
-static const char* find_crlf(const char* p, const char* data, size_t len);
-
 #if defined(_MSC_VER) // SSE2 is baseline
 #include <intrin.h>
 #define __builtin_ctz _tzcnt_u32
@@ -516,7 +515,7 @@ static uint32_t get_crlf_mask(const char* p) {
   return _mm_movemask_epi8(v2);
 }
 
-static const char* find_crlf(const char* p, const char* data, size_t len) {
+const char* find_crlf(const char* p, const char* data, size_t len) {
   const char* lastp = MIN(data + len, HTTP_MAX_HEADER_SIZE + p);
   uint32_t result = 0;
 
@@ -562,7 +561,7 @@ static const char* find_crlf(const char* p, const char* data, size_t len) {
 
 #else
 
-static const char* find_crlf(const char* p, const char* data, size_t len) {
+const char* find_crlf(const char* p, const char* data, size_t len) {
   const char* p_cr;
   const char* p_lf;
   size_t limit = data + len - p;
--- a/http_parser.c
+++ b/http_parser.c
@@ -485,52 +485,45 @@ int http_message_needs_eof(const http_parser *parser);
 
 const char* find_crlf(const char* p, const char* data, size_t len);
 
-
-#if defined(__SSE2__) && defined(__GNUC__) 
-
+#if defined(_MSC_VER) // SSE2 is baseline
+#include <intrin.h>
+#define __builtin_ctz _tzcnt_u32
+#elif defined(__SSE2__) && defined(__GNUC__)
 #include <emmintrin.h>
+#endif
+
+#if defined(_MSC_VER) || (defined(__SSE2__) && defined(__GNUC__))
+
+int32_t get_crlf_mask(const char* p) {
+  /* [ c, 0, 0, 0, 0, 0 .. 0 ] */
+  const __m128i vCR = _mm_set1_epi8(0x0a);
+
+  /* [ c, 0, 0, 0, 0, 0 .. 0 ] */
+  const __m128i vLF = _mm_set1_epi8(0x0d);
+
+  __m128i v1, v2;
+  v1 = *(__m128i*)p;
+  v2 = _mm_cmpeq_epi8(vCR, v1);
+  v1 = _mm_cmpeq_epi8(vLF, v1);
+  v2 = _mm_or_si128(v1, v2);
+  return _mm_movemask_epi8(v2);
+}
 
 const char* find_crlf(const char* p, const char* data, size_t len) {
   const char* lastp = MIN(data + len, HTTP_MAX_HEADER_SIZE + p);
 
   int32_t result = 0;
-  __m128i v1, v2;
-  /* [ c, 0, 0, 0, 0, 0 .. 0 ] */
-  __m128i vCR = _mm_set_epi32(0x0a0a0a0a, 0x0a0a0a0a,0x0a0a0a0a, 0x0a0a0a0a);
-
-  /* [ c, 0, 0, 0, 0, 0 .. 0 ] */
-  __m128i vLF = _mm_set_epi32(0x0d0d0d0d, 0x0d0d0d0d,0x0d0d0d0d, 0x0d0d0d0d);
 
   size_t alignment = (long) p & 15;
   p -= alignment;
 
-  v1 = *(__m128i*)p;
-  v2 = _mm_cmpeq_epi8(vCR, v1);
-  v1 = _mm_cmpeq_epi8(vLF, v1);
-  v2 = _mm_or_si128(v1, v2);
-  result = _mm_movemask_epi8(v2);
-
-  v1 = *(__m128i*)(p + 16);
-  v2 = _mm_cmpeq_epi8(vCR, v1);
-  v1 = _mm_cmpeq_epi8(vLF, v1);
-  v2 = _mm_or_si128(v1, v2);
-  result = _mm_movemask_epi8(v2) << 16 | result;
+  result = get_crlf_mask(p + 16) << 16 | get_crlf_mask(p);
   result = (result >> alignment) << alignment;
 
   if (!result) {
     while(!result && lastp >= p+32) {
       p += 32;
-      v1 = *(__m128i*)p;
-      v2 = _mm_cmpeq_epi8(vCR, v1);
-      v1 = _mm_cmpeq_epi8(vLF, v1);
-      v2 = _mm_or_si128(v1, v2);
-      result = _mm_movemask_epi8(v2);
-
-      v1 = *(__m128i*)(p+16);
-      v2 = _mm_cmpeq_epi8(vCR, v1);
-      v1 = _mm_cmpeq_epi8(vLF, v1);
-      v2 = _mm_or_si128(v1, v2);
-      result = _mm_movemask_epi8(v2) << 16 | result;
+	  result = get_crlf_mask(p + 16) << 16 | get_crlf_mask(p);
     }
     if (!result) {
       return data + len;
--- a/http_parser.c
+++ b/http_parser.c
@@ -444,7 +444,7 @@ enum http_host_state
     }                                                         \
     ch = *++p;                                                \
     ++parser->nread;                                          \
-  } else                                                      \
+  } else {}                                                   \
 
 
 /**
--- a/http_parser.c
+++ b/http_parser.c
@@ -500,10 +500,10 @@ const char* find_crlf(const char* p, const char* data, size_t len);
 
 int32_t get_crlf_mask(const char* p) {
   /* [ c, 0, 0, 0, 0, 0 .. 0 ] */
-  const __m128i vCR = _mm_set1_epi8(0x0a);
+  const __m128i vCR = _mm_set1_epi8(0x0d);
 
   /* [ c, 0, 0, 0, 0, 0 .. 0 ] */
-  const __m128i vLF = _mm_set1_epi8(0x0d);
+  const __m128i vLF = _mm_set1_epi8(0x0a);
 
   __m128i v1, v2;
   v1 = *(__m128i*)p;
--- a/http_parser.c
+++ b/http_parser.c
@@ -435,13 +435,15 @@ enum http_host_state
   (IS_ALPHANUM(c) || (c) == '.' || (c) == '-' || (c) == '_')
 #endif
 
-#define NEXTHEADERCHAR() \
-  if (UNLIKELY(p+1 == (data+len))) {                          \
-    COUNT_HEADER_SIZE(0);                                     \
-    break;                                                    \
-  }                                                           \
-  ch = *++p;                                                  \
-  ++parser->nread;
+#define NEXTHEADERCHAR()                                      \
+  if (1) {                                                    \
+    if (UNLIKELY(p + 1 == data + len)) {                      \
+      COUNT_HEADER_SIZE(0);                                   \
+      break;                                                  \
+    }                                                         \
+    ch = *++p;                                                \
+    ++parser->nread;                                          \
+  } else                                                      \
 
 
 /**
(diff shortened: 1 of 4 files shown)


shekhei:
I am not sure how to ignore whitespace when you view the diff, the changes are actually very minimal

craig65535:
@shekhei you can use https://github.com/nodejs/http-parser/pull/342/files?w=1 to see the diff without whitespace changes.

BTW it looks like the style in this file is to use lower-case for labels, not upper-case as you have for HEADER_FIELD_BEGIN

indutny:
@shekhei thank you for doing this! I've experimented with vector intrinsics some time before, but for whatever reason I had at the moment decided to not proceed further.

Is the most of the speedup achieved by using vector intrinsics? How much do other changes contribute?

shekhei:
It seems like they each contribute to approximately half of the speed up. The fallthroughs about 10%, the h_general jmp was about another 20%, the vector speedups is slightly more interesting, it really depends on the size of the header values :)

I was really surprised about the gcc and clang difference though, if i use gcc 6 to compile it, my god its almost twice as fast as the clang one on osx.

shekhei:
@craig65535 you are right, let me modify them to lower case, and using snake for function names.

shekhei:
hmm.. any news on this pr?

http_parser.c
```diff
@@ -472,6 +481,88 @@ static struct {
...
+#if defined(__SSE2__) && defined(__GNUC__) 
+
+#include <emmintrin.h>
+
+const char* find_crlf(const char* p, const char* data, size_t len) {
+  const char* lastp = MIN(data+len, HTTP_MAX_HEADER_SIZE+p);
+
+  int32_t result = 0;
+  __m128i v1, v2;
+  __m128i vCR = _mm_set_epi32(0x0a0a0a0a, 0x0a0a0a0a,0x0a0a0a0a, 0x0a0a0a0a); // [ c, 0, 0, 0, 0, 0 .. 0 ]
```
indutny: Please use C style comments `/* */`

indutny: Also 80 column limit?

http_parser.c
```diff
@@ -472,6 +481,88 @@ static struct {
...
+  result = _mm_movemask_epi8(v2);
+
+  v1 = *((__m128i*)(p + 16));
+  v2 = _mm_cmpeq_epi8(vCR, v1);
+  v1 = _mm_cmpeq_epi8(vLF, v1);
+  v2 = _mm_or_si128(v1, v2);
+  result = (_mm_movemask_epi8(v2) << 16 ) | result;
+  result = (result >> alignment) << alignment;
+
+  if ( !result ) {
```
indutny: `(!result)`

http_parser.c
```diff
@@ -472,6 +481,88 @@ static struct {
...
+  int32_t result = 0;
+  __m128i v1, v2;
+  __m128i vCR = _mm_set_epi32(0x0a0a0a0a, 0x0a0a0a0a,0x0a0a0a0a, 0x0a0a0a0a); // [ c, 0, 0, 0, 0, 0 .. 0 ]
+
+  __m128i vLF = _mm_set_epi32(0x0d0d0d0d, 0x0d0d0d0d,0x0d0d0d0d, 0x0d0d0d0d); // [ c, 0, 0, 0, 0, 0 .. 0 ]
+
+  size_t alignment = (long)p & 15;
+  p -= alignment;
+
+  v1 = *((__m128i*)(p));
```
indutny: Just `*(__m128i*) p`, maybe?

http_parser.c
```diff
@@ -472,6 +481,88 @@ static struct {
...
+  v2 = _mm_cmpeq_epi8(vCR, v1);
+  v1 = _mm_cmpeq_epi8(vLF, v1);
+  v2 = _mm_or_si128(v1, v2);
+  result = _mm_movemask_epi8(v2);
+
+  v1 = *((__m128i*)(p + 16));
+  v2 = _mm_cmpeq_epi8(vCR, v1);
+  v1 = _mm_cmpeq_epi8(vLF, v1);
+  v2 = _mm_or_si128(v1, v2);
+  result = (_mm_movemask_epi8(v2) << 16 ) | result;
```
indutny: No need in parens.

http_parser.c
```diff
@@ -472,6 +481,88 @@ static struct {
...
+  result = (result >> alignment) << alignment;
+
+  if ( !result ) {
+    while( !result && lastp >= (p+32) ) {
```
indutny: `while (!result && lastp >= p+32) {`

http_parser.c
```diff
@@ -472,6 +481,88 @@ static struct {
...
+      v2 = _mm_or_si128(v1, v2);
+      result = _mm_movemask_epi8(v2);
+
+      v1 = *((__m128i*)(p+16));
+      v2 = _mm_cmpeq_epi8(vCR, v1);
+      v1 = _mm_cmpeq_epi8(vLF, v1);
+      v2 = _mm_or_si128(v1, v2);
+      result = (_mm_movemask_epi8(v2) << 16 ) | result;
+    }
+    if ( !result ) { return data+len; }
```
indutny: Please put block's statement on a next line.

http_parser.c
```diff
@@ -472,6 +481,88 @@ static struct {
...
+  }
+  p += __builtin_ctz(result);
+  if ( p >= lastp ) {
```
indutny: Please remove spaces after `(` and before `)`.

http_parser.c
```diff
@@ -472,6 +481,88 @@ static struct {
...
+    return data+len;
```
indutny: `data + len`

http_parser.c
```diff
@@ -1292,121 +1383,133 @@ size_t http_parser_execute (http_parser *parser,
...
+                if (ch != ' ') parser->header_state = h_general;
+                break;
+
+              default:
+                assert(0 && "Unknown header_state");
+                break;
+            }
           }
         }
-
```
indutny: Superfluous change?

http_parser.c
```diff
@@ -435,6 +435,15 @@ enum http_host_state
   (IS_ALPHANUM(c) || (c) == '.' || (c) == '-' || (c) == '_')
 #endif
 
+#define NEXTHEADERCHAR() \
+  if (UNLIKELY(p+1 == (data+len))) {                          \
```
indutny: `p + 1 == data + len`

http_parser.c
```diff
@@ -435,6 +435,15 @@ enum http_host_state
   (IS_ALPHANUM(c) || (c) == '.' || (c) == '-' || (c) == '_')
 #endif
 
+#define NEXTHEADERCHAR() \
```
indutny: Please wrap contents of define with `do { } while (0)`, and align `\`.

http_parser.c
```diff
@@ -472,6 +481,88 @@ static struct {
...
+const char* find_crlf(const char* p, const char* data, size_t len) {
+  const char* lastp = MIN(data+len, HTTP_MAX_HEADER_SIZE+p);
+
+  int32_t result = 0;
+  __m128i v1, v2;
+  __m128i vCR = _mm_set_epi32(0x0a0a0a0a, 0x0a0a0a0a,0x0a0a0a0a, 0x0a0a0a0a); // [ c, 0, 0, 0, 0, 0 .. 0 ]
+
+  __m128i vLF = _mm_set_epi32(0x0d0d0d0d, 0x0d0d0d0d,0x0d0d0d0d, 0x0d0d0d0d); // [ c, 0, 0, 0, 0, 0 .. 0 ]
+
+  size_t alignment = (long)p & 15;
```
indutny: `(long) p & 15`

indutny (changes requested):
Thank you so much for this, and sorry for delay! May I ask you to correct the style of the patch before further review?

http_parser.c
shekhei: parens in terms of removing this parens?
```
result = _mm_movemask_epi8(v2) << 16 | result
```

indutny: Yep, they are not required.

http_parser.c
shekhei: hmm... wrapping it around with a while loop while keeping the break is rather tricky, any good tricks for this?

indutny: Oh right, I didn't see break here... It can be left as it is then, sorry!

shekhei: is it ok to use a if (1) { } else instead?

http_parser.c
```diff
@@ -435,6 +435,17 @@ enum http_host_state
   (IS_ALPHANUM(c) || (c) == '.' || (c) == '-' || (c) == '_')
 #endif
 
+#define NEXTHEADERCHAR()                                      \
+  if (1) {                                                    \
```
shekhei: Using if(1) {} else instead because of the break within this macro

http_parser.c
```diff
@@ -472,6 +483,92 @@ static struct {
...
+
+#include <emmintrin.h>
+
+const char* find_crlf(const char* p, const char* data, size_t len) {
+  const char* lastp = MIN(data + len, HTTP_MAX_HEADER_SIZE + p);
+
+  int32_t result = 0;
+  __m128i v1, v2;
+  /* [ c, 0, 0, 0, 0, 0 .. 0 ] */
+  __m128i vCR = _mm_set_epi32(0x0a0a0a0a, 0x0a0a0a0a,0x0a0a0a0a, 0x0a0a0a0a);
```
shekhei: This is within 80col

shekhei:
hmm... :P

shekhei:
hmm... any news?

http_parser.c
```diff
@@ -472,6 +483,92 @@ static struct {
 
 int http_message_needs_eof(const http_parser *parser);
 
+const char* find_crlf(const char* p, const char* data, size_t len);
+
+
+#if defined(__SSE2__) && defined(__GNUC__) 
+
+#include <emmintrin.h>
```
zbjornson: Nice work with this PR!

It's trivial to make this work on Windows also, just change these two lines to:

```cpp
#if defined(_MSC_VER) // SSE2 is baseline
#include <intrin.h>
#define __builtin_ctz _tzcnt_u32
#elif defined(__SSE2__) && defined(__GNUC__) 
#include <emmintrin.h>
#endif

#if defined(_MSC_VER) || (defined(__SSE2__) && defined(__GNUC__))
```

http_parser.c
zbjornson: `_mm_set1_epi32(0x0a0a0a0a)` does the same thing and compiles to one fewer instruction:

```asm
  vmovdqa xmm0, XMMWORD PTR .LC0[rip]
  vmovaps XMMWORD PTR vCR[rip], xmm0
```
vs just:
```asm
  vmovaps XMMWORD PTR vCR[rip], xmm0
```

http_parser.c
```diff
@@ -472,6 +483,92 @@ static struct {
...
+      v2 = _mm_cmpeq_epi8(vCR, v1);
+      v1 = _mm_cmpeq_epi8(vLF, v1);
+      v2 = _mm_or_si128(v1, v2);
+      result = _mm_movemask_epi8(v2);
+
+      v1 = *(__m128i*)(p+16);
+      v2 = _mm_cmpeq_epi8(vCR, v1);
+      v1 = _mm_cmpeq_epi8(vLF, v1);
+      v2 = _mm_or_si128(v1, v2);
+      result = _mm_movemask_epi8(v2) << 16 | result;
```
zbjornson: These four lines are repeated four times; you could add a little `inline static int get_crlf_mask(const char*)` method maybe.

zbjornson:
Eager to see this landed. I get about a 21% speedup on an i7-7700HQ on Windows.

Just tested a wider version (256 bits, requiring AVX2) and get another 8% on top of that. This SSE2-only version needs no dispatching though, which makes it easy. Happy to help with an AVX2 implementation (using either preprocessor directives or dispatching) if there's interest, though.

http_parser.c
shekhei: Thanks for the comments! i made most of the changes and tested, they are all working, as for the inline function, I am not quite sure if that works on every platform's compiler's c dialect though

zbjornson: Awesome! Thanks. (Did you push the changes up?)

As for `inline static`, I saw the line below saying it's a C99 project, which means it's in the standard. GCC has supported it since 4.2. Obviously unnecessary, just nice cleanup.
https://github.com/nodejs/http-parser/blob/master/http_parser.gyp#L65

shekhei: nope, I have yet to push the change, will be doing that tonight( asia time, working hours :P )

Oh what I kinda worry about is that whether the other compilable platforms has the dialect turned on, cause this is my first contribution I am not quite sure, if you are sure about it, let me make that change too!

shekhei: Or perhaps anybody else who might be able to shed a light on whether this is safe to use, or should I write a macro instead?

2underscores-vic: Why don't make a regular function? Will it be a big overhead?

zbjornson: > whether the other compilable platforms has the dialect turned on

GCC and Clang default to gnu11, which supports `inline`. MSVC is set to compile this package as C++. Since you're testing for `__GNUC__`, I think that means it would be safe. But... in any case, I just tested GCC 4.4.7, 5.1, 6.1 and 7.2, and LLVM Clang 4.0.0 and 5.0.0, and they all automatically inline the static helper function without the `inline` keyword in this case.

shekhei: Yup, I just did some test and compiled them last night too, most of them seem to inline it automatically anyways.

shekhei:
Ok, it seems like my whole branch has gone cranky after not updating for so long, let me clean it up

shekhei:
Ok, latest changes made

zbjornson:
@indutny What do you think of landing this awesome PR? Will be nice to have this core part of node.js sped up!

indutny:
This PR is awesome indeed!

I'll try to find time to review it this week. Sorry for not doing it sooner?

@bnoordhuis @mscdex your help would be very appreciated.

http_parser.c
```diff
@@ -472,6 +483,85 @@ static struct {
...
+const char* find_crlf(const char* p, const char* data, size_t len);
+
+#if defined(_MSC_VER) // SSE2 is baseline
+#include <intrin.h>
+#define __builtin_ctz _tzcnt_u32
+#elif defined(__SSE2__) && defined(__GNUC__)
+#include <emmintrin.h>
+#endif
+
+#if defined(_MSC_VER) || (defined(__SSE2__) && defined(__GNUC__))
```
mscdex: Perhaps it might be a good idea to make this an overridable define so that the plain C versions of functions can be tested more easily.

http_parser.c
```diff
@@ -472,6 +483,85 @@ static struct {
...
+  int32_t result = 0;
+
+  size_t alignment = (long) p & 15;
+  p -= alignment;
+
+  result = get_crlf_mask(p + 16) << 16 | get_crlf_mask(p);
+  result = (result >> alignment) << alignment;
+
+  if (!result) {
+    while(!result && lastp >= p+32) {
```
mscdex: Inconsistent spacing. Personally I prefer a space between the keyword and parenthesis.

http_parser.c
```diff
@@ -472,6 +483,85 @@ static struct {
...
+      p += 32;
+	  result = get_crlf_mask(p + 16) << 16 | get_crlf_mask(p);
```
mscdex: Indentation off here.

mscdex:
It would be interesting to see how much of a difference there would be in using NEON on ARM for the parts using intrinsics. I might try that this weekend.

http_parser.c
zbjornson: Using `-mno-sse2` (regardless of `-march` or `-mtune`) would achieve this, perhaps less explicitly than you're aiming for though.

indutny: I agree that overridable define should be introduced here.

http_parser.c
shekhei: done

http_parser.c
shekhei: done, i hope this is what you guys are looking for :P

Kronuz:
Clang doesn’t define `__GNUC__`, yet it does support `emmintrin.h`, please we need to remove the `__GNUC__` checks and just leave `__SSE2__` to make this code efficient also with clang.

zbjornson:
Clang does define `__GNUC__` actually. (Have you found some version that doesn't?)

Kronuz:
Well, here it wasn’t compiling that part and it was because of that `__GNUC__` check. I'll dig deeper.

zbjornson:
Did you have `-march=[something with at least SSE2` or `-msse2`?

http_parser.c
```diff
@@ -472,6 +483,89 @@ static struct {
...
+
+#ifndef USE_INTRISICS_CRLF
+# define USE_INTRISICS_CRLF defined(_MSC_VER) || (defined(__SSE2__) && defined(__GNUC__))
+#endif
+
+#if USE_INTRISICS_CRLF
+
+int32_t get_crlf_mask(const char* p) {
+  /* [ c, 0, 0, 0, 0, 0 .. 0 ] */
+  const __m128i vCR = _mm_set1_epi8(0x0a);
```
mscdex: Shouldn't this be `vLF` and the other `vCR` since 0x0a is line feed?

http_parser.c
```diff
@@ -472,6 +483,89 @@ static struct {
...
+
+const char* find_crlf(const char* p, const char* data, size_t len) {
+  const char* lastp = MIN(data + len, HTTP_MAX_HEADER_SIZE + p);
+
+  int32_t result = 0;
+
+  size_t alignment = (long) p & 15;
+  p -= alignment;
+
+  result = get_crlf_mask(p + 16) << 16 | get_crlf_mask(p);
```
mscdex: Don't we need to check that `len >= 16` before trying to use the intrinsics which operate on 16 bytes at a time? For example, a valid HTTP request could just be `GET /<CR><LF><CR><LF>` which is only 9 bytes.

http_parser.c
mscdex: I get multiple compiler errors for this same macro, looks like additional braces are needed?:

```
http_parser.c:937:25: error: suggest braces around empty body in an ‘else’ statement [-Werror=empty-body]
         NEXTHEADERCHAR();
```

mscdex:
Ok I just added a NEON-based implementation and I see ~13% improvement when testing on an Odroid C2:

Before:

```
Benchmark result:
Took 29.264072 seconds to run
170857.968750 req/sec
```

After:

```
Benchmark result:
Took 25.583895 seconds to run
195435.453125 req/sec
```

mscdex:
FWIW here is my patch on top of this PR as it currently stands:

```diff
--- http_parser.c	2017-09-23 16:50:15.330221304 -0400
+++ http_parser-neon.c	2017-09-23 16:52:40.372395441 -0400
@@ -443,7 +443,7 @@
     }                                                         \
     ch = *++p;                                                \
     ++parser->nread;                                          \
-  } else                                                      \
+  }
 
 
 /**
@@ -485,19 +485,17 @@
 
 const char* find_crlf(const char* p, const char* data, size_t len);
 
+#ifndef USE_INTRISICS_CRLF
+# define USE_INTRISICS_CRLF 1
+#endif
+
+#if USE_INTRISICS_CRLF && (defined(_MSC_VER) || (defined(__SSE2__) && defined(__GNUC__)))
 #if defined(_MSC_VER) // SSE2 is baseline
 #include <intrin.h>
 #define __builtin_ctz _tzcnt_u32
 #elif defined(__SSE2__) && defined(__GNUC__)
 #include <emmintrin.h>
 #endif
-
-#ifndef USE_INTRISICS_CRLF
-# define USE_INTRISICS_CRLF defined(_MSC_VER) || (defined(__SSE2__) && defined(__GNUC__))
-#endif
-
-#if USE_INTRISICS_CRLF
-
 int32_t get_crlf_mask(const char* p) {
   /* [ c, 0, 0, 0, 0, 0 .. 0 ] */
   const __m128i vCR = _mm_set1_epi8(0x0a);
@@ -539,9 +537,63 @@
   }
   return p;
 }
+#elif USE_INTRISICS_CRLF && defined(__ARM_NEON)
+#include <arm_neon.h>
+const uint8_t __attribute__ ((aligned (16))) _powers_array[16] =
+  { 1, 2, 4, 8, 16, 32, 64, 128, 1, 2, 4, 8, 16, 32, 64, 128 };
+uint8x16_t _powers;
+uint8_t _powers_init = 0;
+uint16_t get_crlf_mask(const char* p) {
+  /* [ c, 0, 0, 0, 0, 0 .. 0 ] */
+  const uint8x16_t vCR = vdupq_n_u8(0x0d);
 
-#else
+  /* [ c, 0, 0, 0, 0, 0 .. 0 ] */
+  const uint8x16_t vLF = vdupq_n_u8(0x0a);
+
+  uint8x16_t v1, v2;
+  v1 = *(uint8x16_t*)p;
+  v2 = vceqq_u8(vCR, v1);
+  v1 = vceqq_u8(vLF, v1);
+  v2 = vorrq_u8(v1, v2);
+
+  // Compute the mask from the input
+  uint64x2_t mask = vpaddlq_u32(vpaddlq_u16(vpaddlq_u8(vandq_u8(v2, _powers))));
+
+  // Get the resulting bytes
+  uint16_t output;
+  vst1q_lane_u8((uint8_t*)&output + 0, (uint8x16_t)mask, 0);
+  vst1q_lane_u8((uint8_t*)&output + 1, (uint8x16_t)mask, 8);
+
+  return output;
+}
+
+const char* find_crlf(const char* p, const char* data, size_t len) {
+  const char* lastp = MIN(data + len, HTTP_MAX_HEADER_SIZE + p);
+
+  int32_t result = 0;
+
+  size_t alignment = (long) p & 15;
+  p -= alignment;
+
+  result = get_crlf_mask(p + 16) << 16 | get_crlf_mask(p);
+  result = (result >> alignment) << alignment;
 
+  if (!result) {
+    while (!result && lastp >= p + 32) {
+      p += 32;
+      result = get_crlf_mask(p + 16) << 16 | get_crlf_mask(p);
+    }
+    if (!result) {
+      return data + len;
+    }
+  }
+  p += __builtin_ctz(result);
+  if (p >= lastp) {
+    return data + len;
+  }
+  return p;
+}
+#else
 const char* find_crlf(const char* p, const char* data, size_t len) {
   const char* p_cr;
   const char* p_lf;
@@ -563,7 +615,6 @@
   }
   return p;
 }
-
 #endif
 
 /* Our URL parser.
@@ -2181,6 +2232,13 @@
   parser->type = t;
   parser->state = (t == HTTP_REQUEST ? s_start_req : (t == HTTP_RESPONSE ? s_start_res : s_start_req_or_res));
   parser->http_errno = HPE_OK;
+#if USE_INTRISICS_CRLF && defined(__ARM_NEON)
+  // Only create the powers list once
+  if (_powers_init == 0) {
+    _powers = vld1q_u8(_powers_array);
+    _powers_init = 1;
+  }
+#endif
 }
 
 void
```

shekhei:
@Kronuz actually I have tested it on clang too, it works, maybe you can share which version you are using?

http_parser.c
shekhei: Actually unless there are some safety or memory concerns, then no, the logic is as such

##case1
data: \r\nGET /\r\n\r\n
p: data+2
len: 9
and imagine if alignment is off by 1

*get alignment offset
*p -= 1, which is data+1
*checks next 32 bytes
*result will be that looks something like 1000 0011 1100 1010 1010 1010 1010 1010(it should be right to left, i wrote it left to write)
**anything beyond the 10th bit is for some data that's beyond the range
*shift right and left by alignment offset, so it will end up being 0000 0011 1100 1010 1010 1010 1010 1010
*and so the first bit found would be the first \r from data+2

##case2

data: \r\nGET /\r\n\r\n
p: data+2
len: 4
and imagine if alignment is off by 1

*exactly the same as before
*but the position of the first found \r after removing alignment false positives will be beyond "lastp"
**lastp is either data+len or max header size + p

http_parser.c
shekhei: ah yes, fixed

mscdex:
I've updated my patch a bit.

I'm also wondering if the fallback code might benefit from a SWAR implementation over the current dual `memchr()` solution?

shekhei:
@mscdex I did try optimizing with packing into some 64bit and then comparing them, I don't think the improvements were significant, also given that memchr is always optimized for a specific platform, if my memory serves me right, it started losing out rather quickly.

I might be wrong though.

http_parser.c
bnoordhuis: Macros with control flow in them that don't announce it in their name are a special kind of evil.

http_parser.c
```diff
@@ -472,6 +483,89 @@ static struct {
 
 int http_message_needs_eof(const http_parser *parser);
 
+const char* find_crlf(const char* p, const char* data, size_t len);
```
bnoordhuis: `static`

http_parser.c
```diff
@@ -472,6 +483,89 @@ static struct {
...
+
+#if defined(_MSC_VER) // SSE2 is baseline
+#include <intrin.h>
+#define __builtin_ctz _tzcnt_u32
+#elif defined(__SSE2__) && defined(__GNUC__)
+#include <emmintrin.h>
+#endif
+
+#ifndef USE_INTRISICS_CRLF
+# define USE_INTRISICS_CRLF defined(_MSC_VER) || (defined(__SSE2__) && defined(__GNUC__))
```
bnoordhuis: Long line, please wrap it at 80 columns.  (Also applies to any other long lines this PR may introduce, I won't point them out individually.)

As well, this would be better written as:
```c
#if defined(_MSC_VER) || (defined(__SSE2__) && defined(__GNUC__))
# define USE_INTRISICS_CRLF 1
#endif
```

http_parser.c
```diff
@@ -472,6 +483,89 @@ static struct {
...
+#endif
+
+#if USE_INTRISICS_CRLF
+
+int32_t get_crlf_mask(const char* p) {
```
bnoordhuis: `static`.  I'll stop pointing this out from here on.

http_parser.c
```diff
@@ -472,6 +483,89 @@ static struct {
...
+  /* [ c, 0, 0, 0, 0, 0 .. 0 ] */
+  const __m128i vCR = _mm_set1_epi8(0x0d);
+
+  /* [ c, 0, 0, 0, 0, 0 .. 0 ] */
+  const __m128i vLF = _mm_set1_epi8(0x0a);
+
+  __m128i v1, v2;
+  v1 = *(__m128i*)p;
```
bnoordhuis: Aliases.  It's also not directly obvious to me if it respects alignment.

bnoordhuis (commented):
Left some comments.  I only looked at this really briefly though.

http_parser.c
shekhei: It was modified to this to allow overridable defines, as requested by @mscdex earlier

bnoordhuis: Okay, but defining it like this doesn't work, the `#if USE_INTRISICS_CRLF` is always going to evaluate to true.  Try changing it to `#define USE_INTRISICS_CRLF defined(florp)` and you'll see it still evaluates to true.

http_parser.c
shekhei: Hmm... any recommendations on how would you prefer this to be done for the alignment part? The alignment is ensured in find_crlf and it is done just once before the loop. Should I document it or write it in some comments or do you prefer if I move the alignment code in here? I guess that would incur some unnecessary alignment operations though

http_parser.c
```diff
@@ -472,6 +483,89 @@ static struct {
...
+  v2 = _mm_cmpeq_epi8(vCR, v1);
+  v1 = _mm_cmpeq_epi8(vLF, v1);
+  v2 = _mm_or_si128(v1, v2);
+  return _mm_movemask_epi8(v2);
+}
+
+const char* find_crlf(const char* p, const char* data, size_t len) {
+  const char* lastp = MIN(data + len, HTTP_MAX_HEADER_SIZE + p);
```
bnoordhuis: `HTTP_MAX_HEADER_SIZE + p` is not legal C when it points beyond the last element + 1 of the array that `p` points to.  Not a strictly academic concern: whole program optimizers can see through this.

http_parser.c
```diff
@@ -472,6 +483,89 @@ static struct {
...
+
+  int32_t result = 0;
```
bnoordhuis: Shifting into the sign bit (what you do below) is implementation-defined behavior.  This should be `uint32_t`.

http_parser.c
```diff
@@ -472,6 +483,89 @@ static struct {
...
+
+  size_t alignment = (long) p & 15;
```
bnoordhuis: Use `uintptr_t`, `long` isn't necessarily wide enough to hold a pointer.

http_parser.c
```diff
@@ -472,6 +483,89 @@ static struct {
...
+  p -= alignment;
```
bnoordhuis: What guarantees that this is a valid pointer?

http_parser.c
bnoordhuis: It reads beyond the end of the input, doesn't it?  Or am I misunderstanding your comment?

shekhei: Ah, yes it does, this pr's been so long, i think i've lost myself somewhere. Let me get that fixed, it should be changed to matching one by one till aligned and then matching remaining one by one

http_parser.c
zbjornson: (GCC defines the vector types with `__may_alias__`, making this safe. If you're worried about memory alignment you could use `_mm_lddqu_si128` explicitly.)

shekhei:
made some changes to ensure not going before or after requested bounds within findcrlf, did a bench, almost no drop in performance

http_parser.c
shekhei: Made a change to the logic

while misaligned and not last char
  checks char one by one
  returns if found
if remaining is less than 32:
  while not last char
    checks char one by one
    return if found
  return not found
SIMD match till there are at least 32 left, return if found
if not found and not ended yet
  find char one by one and return if found
return if not found

this change will also ensure that we will not be offsetting the pointer to a possibly bogus address, nor will we read beyond the input anymore.

http_parser.c
shekhei: I have changed the current version to looping unaligned bytes and anything less than 32 one by one, maybe you guys can also review the new commit instead? anyways _mm_lddqu_si128 is unfortunately sse3, right now i think we are just using sse2 for baseline

shekhei: @zbjornson I did try it out and loaded the sse3 headers, such a trivial change actually gave me a consistent 4% improvement. Unfortunately I think we might be rather out of luck though.

mscdex: Why out of luck? I don't see a problem with requiring SSE3, which was introduced in early 2004.

shekhei: @mscdex I am not sure what is the "baseline" for the supported architectures, I decided to try making use of sse2 instruction sets only because v8 requires sse2, so I suppose that was the baseline, if we are actually cool with supporting higher versions of vector instructions then great, I will change it immediately! :)

zbjornson: I forgot that `lddqu` is SSE3. Sticking to SSE2 is nice because it's baseline on all platforms, and there's not much else you can do to get a perf bump here until you get to 256-bit registers (AVX) or get `pshufb` (SSSE3).

As far as improving node's performance specifically, node's build would have to be tweaked to emit for that instruction set. None of the oldest [currently supported x86/64 platforms](https://github.com/nodejs/node/blob/master/BUILDING.md#supported-platforms-1) require more than SSE2 AFAIK. It would be nice if node.js was open to requiring minimum CPU features (especially ones older than 8-10 years); there's a lot that could be done to improve performance if that were to happen.

I'm a bit surprised that `lddqu` is that much faster, that should only be the case if you're routinely loading across a CL, and I wouldn't expect this code to be memory-bound overall. (The benchmark I think always uses the same payload, so I guess not surprising.) I'll look in more detail today ... I'm also wondering if doing alignment is beneficial overall given that headers tend to be short.

shekhei: @zbjornson I was rather surprised too, such a simple operation. Actually with what I have modified, so we dont end up reading out of bounds, is basically the same as how memchr is implemented with sse2, but i guess its good enough.

I actually tried a avx2 version, with ymm registers and the performance improvements wasnt great, rather surprising. I tested it on a mid 2014 mbp 15" with haswell family i7, maybe things gets faster on the newer ones, not too sure.

shekhei: @zbjornson by the way I think headers can get rather large, think about cookies, i read somewhere typical http headers are between 200bytes to 2kb nowadays.

http_parser.c
```diff
@@ -472,6 +483,109 @@ static struct {
 
 int http_message_needs_eof(const http_parser *parser);
 
+static const char* find_crlf(const char* p, const char* data, size_t len);
```
indutny: I have a proposal. What do you think about moving this into a separate header file? This way we could write extensive tests for this function.

indutny (approved):
Just one suggestion and (implied) request for tests, otherwise LGTM.

http_parser.c
shekhei: so do you mean that we create another header file or do you just mean moving it to http_parser.h?

Or could we perhaps have a non-static function that is conditionally included based on some flags that calls this static function? if not this kinda gets exposed unnecessarily.

indutny: I'd just put it into internal header file, so that it could be included and covered by tests.

shekhei: ok, got it, let me do that in a while.

http_parser.c
zbjornson: As you have it now, I'm not seeing a difference between `lddqu` and `movdqa` (I don't think `lddqu` is "supposed" to be faster than `movdqa` in the case of CL splits with 16B-alignment).

As for AVX2, I get a 7-8% speedup over the current SSE2 version, and appears to be bottlenecking on branches. (`find_crlf` is down to <5% of the total instructions retired though. `parse_url_char` is the next heaviest hitter, which is also unsurprisingly bottlenecked on branches. There's maybe some stuff to be done with LUTs or `pshufb` in that function, for another time.)

shekhei: @zbjornson yup the branches is a huge bottleneck, the function is just so large.

Some of the stuff I've added that shaved off about 5%(not the SIMD part) was actually reverse engineered from clang's compiled version, and then converted back to c, gcc(i didnt look at the asm from the latest gcc though, was using gcc6 i think) seems to be rather weak when it comes to optimizing certain branches.

But definitely did a much better job at optimizing the large function as a whole compared to clang(cant remember the version either)

I actually wrote one more version, just to test it out using PCMPESTRI, it was significantly slower, god knows why :P

shekhei:
hmm... I have added some tests a while ago, any comments? or anything else that I should change?

mscdex:
@shekhei It looks like there was a [build error](https://travis-ci.org/nodejs/http-parser/jobs/280920471) during the last CI run.

shekhei:
Ah, it seems like it is stricter with gcc, i have modified, lets see how's the build

http_parser.c
```diff
@@ -472,6 +484,107 @@ static struct {
...
+#define __builtin_ctz _tzcnt_u32
+#elif defined(__SSE2__) && defined(__GNUC__)
+#include <emmintrin.h>
+#endif
+
+#ifndef USE_INTRISICS_CRLF
+# if defined(_MSC_VER) || \
+    (defined(__SSE2__) && defined(__GNUC__))
+#   define USE_INTRISICS_CRLF 1
+# endif
```
mscdex: This is probably a rarity, but perhaps we should put an `else` here and set the define to 0.

http_parser.c
```diff
@@ -472,6 +484,107 @@ static struct {
...
+  v2 = _mm_or_si128(v1, v2);
+  return _mm_movemask_epi8(v2);
+}
+
+const char* find_crlf(const char* p, const char* data, size_t len) {
+  const char* lastp = MIN(data + len, HTTP_MAX_HEADER_SIZE + p);
+  uint32_t result = 0;
+
+  while ((uintptr_t)p & 15 && p <= lastp) {
+    if ( *p == CR || *p == LF ) {
```
mscdex: The extra spacing like this should be removed to match the style of the rest of the code.

test.c
```diff
@@ -3377,6 +3378,125 @@ test_parse_url (void)
...
+  for (i = 0; i < (sizeof(find_crlf_tests) / sizeof(find_crlf_tests[0])); i++) {
+    test = &find_crlf_tests[i];
+    const char* n = find_crlf(test->p + test->offset, test->p, test->len);
+    if (!test->should_find) {
+      if (n != test->p + test->len) {
+        printf("test_find_crlf(%s) should not find any crlf\n", test->name);
+        printf("test should end at %d instead it is at %ld\n", test->len, n - test->p);
+        abort();
+      }
+    } else if (n != test->p + test->position){
```
mscdex: Minor nit: missing space before brace.

test.c
```diff
@@ -3377,6 +3378,125 @@ test_parse_url (void)
...
+  }
+  ,{.name="long string with crlf"
+  ,.p="                                               \n"  /* 16*3 */
+  ,.offset=0
+  ,.len=48
+  ,.should_find=1
+  ,.position=47
+  }
+  ,{.name="short string without crlf"
+  ,.p="  "  /* 16*3 */
```
mscdex: What is the significance of this same code comment for each newly added test?

shekhei: nothing significant, just a badly done copy and paste :P

shekhei:
Ok made some changes according to comments

mscdex:
Still green. LGTM if the suggestion made by @bnoordhuis about `HTTP_MAX_HEADER_SIZE + p` is not a blocker.

http_parser.c
shekhei: hmm... any recommendations on how this could be improved to be more legal?

shekhei:
hmmmmm.... any news?

shekhei:
@bnoordhuis hmm.. I wonder how is this going? its been one and a half year now :P

bnoordhuis:
There were some instances of UB last time I looked.  Have those been fixed?

I'm curious if there are actual speedups in larger programs.  Mixing SIMD and non-SIMD code sometimes performs worse than just non-SIMD code.

shekhei:
I believe all are fixed other then the
```
const char* lastp = MIN(data + len, HTTP_MAX_HEADER_SIZE + p);
```
which you've mentioned "Not a strictly academic concern"

On larger programs, that would be a good question though, let me try and get hold of a slightly larger web application and test them out.

bnoordhuis:
UB is not something you can try out and see if it works.  It needs to be fixed properly.

shekhei:
no, not the UB part, but the large program performance.

The UB parts were all modified and fixed, other then the one mentioned which was "not a strictly academic concern", the rest were memory bound problems that are fixed, perhaps you can take a look at the latest code commits?

PR #347, opened 2016-12-29
parser: fix handling of newlines before HTTP responses

shindo:
This diff fixes ignoring of CR and LF chars received before HTTP response's start line the same way as they are handled with HTTP_REQUEST and HTTP_BOTH parser types.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -757,16 +757,14 @@ reexecute:
 
       case s_start_res:
       {
+        if (ch == CR || ch == LF)
+          break;
         parser->flags = 0;
         parser->content_length = ULLONG_MAX;
 
         switch (ch) {
           case 'H':
             UPDATE_STATE(s_res_H);
-            break;
-
-          case CR:
-          case LF:
             break;
 
           default:
--- a/test.c
+++ b/test.c
@@ -3891,6 +3891,39 @@ test_message_connect (const struct message *msg)
   parser_free();
 }
 
+/* Verify that `on_message_begin` callback isn't called until the first
+ * byte of message's start line is received. */
+void
+test_message_empty_prefix_lines (enum http_parser_type type)
+{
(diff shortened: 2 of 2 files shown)


indutny:
Hello!

Thank you for submitting this. May I ask you to share your rationale for this change?

As far as I see it - it looks very non-standard. This may be the reason why it is disallowed right now.

shindo:
Well, it's actually allowed right now. Currently if parser is at s_start_res state and receives either CR or LF, it consumes the char, doesn't change its state and calls on_message_begin callback. Thus, if one receives multiple empty lines before response their on_message_begin callback will be called multiple times. Is it intended? 01:45, December 31, 2016, Fedor Indutny <notifications@github.com>:Hello!
Thank you for submitting this. May I ask you to share your rationale for this change?
As far as I see it - it looks very non-standard. This may be the reason why it is disallowed right now.

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.


  
  




Sent from Yandex.Mail for mobile: http://m.ya.ru/ymail

shindo:
Hi!

Once again, this patch is aimed to fix current behaviour of parser's `on_message_begin` callback invocation in case of HTTP_RESPONSE type.

With any parser's type (HTTP_REQUEST, HTTP_RESPONSE, HTTP_BOTH) empty lines between HTTP messages are allowed and ignored. But in case of HTTP_RESPONSE parser's type `on_message_begin` callback is called on each CR/LF char between messages, while it should "silently" ignore these chars.

ploxiln:
Someone could argue that the extra cr/lf chars should be parse errors instead.

shindo:
Sure, but the patch is not about parser's robustness, as currently extra CR/LF chars between messages are already allowed.

PR #348, opened 2017-01-01
Unbreak `make install` in BSDs

roman-neuhauser:
I've only tested this in freebsd-12.0 where the changes get me from syntax errors in the Makefile to passing `port test` with flying colors.


roman-neuhauser:
oh, so this duplicates work done by others multiple times even, all that work hanging in open pull requests for many months in some cases.  c'mon guys!

PR #349, opened 2017-01-08
Add line-by-line comments for http_parser_execute

AoiKuiyuyou:


indutny:
Hello!

Thank you for contributing this. What's the motivation for this change?

AoiKuiyuyou:
Just for readability. Line-by-line comments make the intention behind each line explicit so that future readers can understand the intention easily and read the code fast.

indutny:
I'm afraid that it makes code less readable, because less code fits on the screen. Most of the comments are just copying what code below says, so the provided value is not high enough. Sorry, but I'm going to reject this PR.

AoiKuiyuyou:
No problem. Thanks for handling the PR.

PR #352, opened 2017-02-15
http header parsing

kunalspathak:
http_parser along with other http request information supplies header
fields to the consumer. In `node` case, it creates a new `JSString`
of these header fields which involves memory allocation. If parser
keeps track of headers while parsing and return a unique id to `node`
that help distinguish the known header name, `node` don't have to
allocate for known headers and it can use the pre-created `JSString`
for the header fields. See discussion in https://github.com/nodejs/node/issues/10930

The approach I have taken is while scanning the request, keep track
of known headers and update the `header_state` accordingly. `node` or
any consumer can consume the `header_state` information and decide
to use cached http header field instead of creating new string. I try to
match traditional cases http header. e.g. In Content-Type 'C' and 'T' has
to be upper case for the header to qualify for traditional cased header.
I use `traditional_case_http_headers` to track this information.

Performance is flat for node's http benchmark except bench-parse where
sometimes I noticed 5% regression. I am not sure if it was due to noise.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -162,14 +162,40 @@ do {                                                                 \
 
 
 #define PROXY_CONNECTION "proxy-connection"
-#define CONNECTION "connection"
+#define ACCEPT "accept"
+#define CONTENT_ "content-"
 #define CONTENT_LENGTH "content-length"
-#define TRANSFER_ENCODING "transfer-encoding"
 #define UPGRADE "upgrade"
 #define CHUNKED "chunked"
 #define KEEP_ALIVE "keep-alive"
 #define CLOSE "close"
+#define IF_ "if-"
+#define PROXY_ "proxy-"
+#define X_FORWARDED_ "x-forwarded-"
 
+
+#define case_header(tag, string)                          \
+      case h_matching_##tag:                                    \
+        parser->index++;                                        \
+        if (parser->index > sizeof(string) - 1                  \
+            || c != string[parser->index]) {                    \
+          parser->header_state = h_general;                     \
+        } else if (parser->index == sizeof(string) - 2) {       \
+          parser->header_state = h_##tag;                       \
+        }
+
+#define case_header_with_dash(tag, string)                \
+        case_header(tag, string)                          \
+        else if (c == '-') {                                    \
+            checkForTraditionalCase = true;                     \
+        } else if (checkForTraditionalCase) {                   \
+            parser->traditional_case_http_headers &= (c != ch); \
+            checkForTraditionalCase = false;                    \
+         }
+
+#define default_h_general \
+        default:  \
+        parser->header_state = h_general;
 
 static const char *method_strings[] =
   {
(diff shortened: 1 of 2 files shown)


kunalspathak:
Cc - @jasnell , @indutny

http_parser.c
```diff
@@ -1297,26 +1392,62 @@ size_t http_parser_execute (http_parser *parser,
...
+                case 'c':
+                  parser->header_state = h_complete_accept;
+                  break;
+                case 'g':
+                  parser->header_state = h_matching_age;
+                  break;
+                case 'u':
+                  parser->header_state = h_matching_authorization;
+                  break;
+                default_h_general
```
bnoordhuis: I don't love this macro, I have to say.

http_parser.c
```diff
@@ -1326,78 +1457,294 @@ size_t http_parser_execute (http_parser *parser,
...
+                  parser->header_state = h_EX;
                   break;
+                default_h_general
               }
               break;
 
-            /* connection */
+            case h_EX:
+              parser->index++;
+              parser->header_state = c == 'p' ? h_EXP : h_general;
```
bnoordhuis: Can you wrap in parens?

http_parser.c
```diff
@@ -1326,78 +1457,294 @@ size_t http_parser_execute (http_parser *parser,
...
+                default_h_general
+              }
+              break;
 
-            case h_matching_proxy_connection:
+            case h_R:
               parser->index++;
-              if (parser->index > sizeof(PROXY_CONNECTION)-1
-                  || c != PROXY_CONNECTION[parser->index]) {
+              parser->header_state = c == 'e' ? h_RE : h_general;
```
bnoordhuis: Ditto.  I'll stop pointing it out from now on.

http_parser.c
```diff
@@ -1326,78 +1457,294 @@ size_t http_parser_execute (http_parser *parser,
...
+                    parser->traditional_case_http_headers &= (c != ch);
+                    parser->header_state = h_matching_x_forwarded_proto;
+                    break;
+                    default_h_general
+                }
+              } else if (parser->index > sizeof(X_FORWARDED_) - 1
+                  || c != X_FORWARDED_[parser->index]) {
+                parser->header_state = h_general;
+              } else if (c == '-') {
+                // check traditional casing'F' in X-Forwarded
```
bnoordhuis: Please use C89-style comments, not C99/C++-style comments.

http_parser.h
```diff
@@ -297,6 +344,7 @@ struct http_parser {
   unsigned int header_state : 7; /* enum header_state from http_parser.c */
   unsigned int index : 7;        /* index into current matcher */
   unsigned int lenient_http_headers : 1;
+  unsigned int traditional_case_http_headers : 1;
```
bnoordhuis: I think this changes the size of struct http_parser, which would make it a semver-major change.

indutny:
I believe this should make performance slightly worse, so the posted results are somewhat expected. The function is already gigantic. Compiler has troubles optimizing it.

What does `make bench` report in this repo? What's the difference?

http_parser.c
kunalspathak: Ok, will change it back. But curious, you didn't like the name of macro or you feel macro is too much for this statement?

http_parser.c
```diff
@@ -1297,26 +1392,62 @@ size_t http_parser_execute (http_parser *parser,
 
       case s_header_field:
       {
+        bool checkForTraditionalCase = false;
```
kunalspathak: Changed it.

http_parser.c
```diff
@@ -162,14 +162,40 @@ do {                                                                 \
...
+#define X_FORWARDED_ "x-forwarded-"
+
+
+#define case_header(tag, string)                          \
+      case h_matching_##tag:                                    \
+        parser->index++;                                        \
+        if (parser->index > sizeof(string) - 1                  \
+            || c != string[parser->index]) {                    \
+          parser->header_state = h_general;                     \
+        } else if (parser->index == sizeof(string) - 2) {       \
```
kunalspathak: `sizeof(string) - 2` means that we have just read the last character (and that matched the expected character from `string[parser->index] == c` inside if above). So we are good to save the `header_state`. This is how headers are [matched today](https://github.com/nodejs/http-parser/blob/master/http_parser.c#L1368) and I used same logic to match other headers.

kunalspathak:
@indutny 
> Compiler has troubles optimizing it.

Yes, I was thinking of optimizing this by arranging cases that are more frequent closer to the `switch` statement so we avoid long jumps, but there is no easy way to know which cases would be more frequent. If `C++` compiler optimize `switch/case` with binary search, I can try arranging cases in ascending order of their values.

> What does make bench report in this repo? What's the difference?

I will get these numbers and update this thread.

http_parser.c
bnoordhuis: The latter, it obscures things and it saves only a few keystrokes.

indutny:
@kunalspathak so all in all it does not get much of a performance improvement, and we see some regressions for sure.

kunalspathak:
Yeah, I didn't know that i can test the performance using `make bench`. I will see if i can squeeze out some of the regression.

Issue #356, opened 2017-03-07
[enhancement] set install_name on Darwin platforms


grobian:
Below is  a "trivial" patch :)

```
Makefile: set install_name on Darwin

https://bugs.gentoo.org/show_bug.cgi?id=546098

--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,8 @@
 ifneq (darwin,$(PLATFORM))
 # TODO(bnoordhuis) The native SunOS linker expects -h rather than -soname...
 LDFLAGS_LIB += -Wl,-soname=$(SONAME)
+else
+LDFLAGS_LIB += -Wl,-install_name,$(LIBDIR)/$(SONAME)
 endif

 test: test_g test_fast
```

Note that while I use $(SONAME) for consistency, it feels odd because it includes the full version number.  SONAME/install_name typically is something like libhttp_parser.2.dylib (or libhttp_parser.so.2), referencing the major version component only.  But from the looks of it, this seems intentional, suggesting e.g. 2.6 and 2.7 release are ABI incompatible, as well as 2.6.2 and 2.6.3 would.  (E.g. currently any http-parser update will break all consumers.)

bnoordhuis:
I think a pull request would be welcome.  I'd flip the clauses and use `ifeq (darwin,...)` for clarity.

> Note that while I use $(SONAME) for consistency, it feels odd because it includes the full version number. [..]  suggesting e.g. 2.6 and 2.7 release are ABI incompatible, as well as 2.6.2 and 2.6.3 would.

When I added the soname for #151, I included the minor because minor releases are not backwards compatible with older minor releases (they add new APIs.)  The patch number is not supposed to be in there, though.

cc @indutny since you landed 2896229 which added that.

grobian:
It was introduced in https://github.com/nodejs/http-parser/commit/36f107fa2e62ca1f0fe7087e49f58624c45a24f1 IMO (perhaps using $(MAJ) and $(MIN) helps to avoid this issue from re-appearing)

grobian:
see PR #358 I left the versioning bits aside, but can do a suggestion for that too if you like.

bnoordhuis:
You mean the MAJOR/MINOR thing?  Sounds good.

grobian:
as you wish ;)

PR #358, opened 2017-03-14
Makefile: set install_name on Darwin, issue #356

grobian:
install_name on Mach-O is similar to SONAME on ELF, except that it needs
a full path instead of just the basename.


bnoordhuis:
Thanks Fabian, landed in fd3850c.

PR #359, opened 2017-03-14
Makefile: set versions for SONAME correctly

grobian:
Raised in issue #356, reduce version number in SONAME to MAJOR.MINOR.
While at it, create a symlink the from SONAME to the library, instead of
the other way around, and add a (standard) unversioned symlink to the
library to aid the ordinary linking process.


indutny:
Landed in 8ea9532378291b5a9c27ab61bb5cff336cc396b5, thank you! Sorry for delay!

PR #360, opened 2017-03-15
Include stddef.h instead of sys/types.h.

gatzka:
The include is required for type size_t. stddef.h should be available
on every platform, sys/types.h is not.


indutny:
Landed in 291a32c, thank you!

indutny:
Sorry for very long delay!

Issue #363, opened 2017-04-14
Upgrade header can be used to advertise protocol support


veesahni:
Per RFC7230 (which supercedes 2616):

```
A server MAY send an Upgrade header field in any other response to
advertise that it implements support for upgrading to the listed
protocols, in order of descending preference, when appropriate for a
future request.
```

Accordingly, Apache advertises HTTP2 support with Upgrade headers in normal HTTP responses:

```
HTTP/1.1 200 OK
Date: Fri, 14 Apr 2017 03:08:10 GMT
Server: Apache
Upgrade: h2,h2c
Connection: Upgrade
Content-Length: 26
Content-Type: text/html; charset=utf-8
```

However, for such upgrade advertisements, http-parser thinks an upgrade is in progress and doesn't process the body.

If an Upgrade header is present in a 200 response, it should be ignored.  It's only an upgrade in a 101 response.

bnoordhuis:
#364

ploxiln:
fix was merged

PR #364, opened 2017-04-14
Ignore Upgrade header outside of 101 response.

bnoordhuis:
See #363.  Open question: could anyone be relying on the current behavior?

### DIFF
--- a/test.c
+++ b/test.c
@@ -1771,6 +1771,167 @@ const struct message responses[] =
   ,.chunk_lengths= { 2 }
   }
 
+#define HTTP_101_RESPONSE_WITH_UPGRADE_HEADER 22
+, {.name= "HTTP 101 response with Upgrade header"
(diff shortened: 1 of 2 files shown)


test.c
```diff
@@ -2476,7 +2637,9 @@ message_eq (int index, int connect, const struct message *expected)
     if (!r) return 0;
   }
 
-  MESSAGE_CHECK_STR_EQ(expected, m, upgrade);
+  if (!connect) {
+    MESSAGE_CHECK_STR_EQ(expected, m, upgrade);
+  }
```
bnoordhuis: Test runner bug fix.  This check doesn't make sense for `test_message_connect(&responses[i])` (which is what calls this function) but it went unnoticed because none of the responses had the `.upgrade` field set.

indutny:
Landed in 260c522 and 05525c5. Thank you!

PR #366, opened 2017-05-16
parser: fix HTTP version parsing

mscdex:
Only one digit is allowed for the major version and only one is allowed for the minor version [according to RFC 7230](https://tools.ietf.org/html/rfc7230#appendix-B).

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -286,10 +286,10 @@ enum state
   , s_res_HT
   , s_res_HTT
   , s_res_HTTP
-  , s_res_first_http_major
   , s_res_http_major
-  , s_res_first_http_minor
+  , s_res_http_dot
   , s_res_http_minor
+  , s_res_http_end
   , s_res_first_status_code
   , s_res_status_code
   , s_res_status_start
@@ -316,10 +316,10 @@ enum state
   , s_req_http_HT
   , s_req_http_HTT
   , s_req_http_HTTP
-  , s_req_first_http_major
   , s_req_http_major
-  , s_req_first_http_minor
+  , s_req_http_dot
   , s_req_http_minor
+  , s_req_http_end
   , s_req_line_almost_done
 
   , s_header_field_start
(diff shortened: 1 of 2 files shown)


test.c
```diff
@@ -3313,9 +3313,9 @@ test_message_count_body (const struct message *message)
 }
 
 void
-test_simple (const char *buf, enum http_errno err_expected)
+test_simple_type (const char *buf, enum http_errno err_expected, enum http_parser_type type)
```
indutny: Nitpicking: does it violate 80 column limit?

http_parser.c
```diff
@@ -1217,19 +1175,8 @@ size_t http_parser_execute (http_parser *parser,
 
         /* XXX allow spaces after digit? */
```
indutny: Do we still want to allow any spaces after the digit?

http_parser.c
```diff
@@ -795,75 +795,49 @@ size_t http_parser_execute (http_parser *parser,
...
-        }
-
-        parser->http_major *= 10;
-        parser->http_major += ch - '0';
-
-        if (UNLIKELY(parser->http_major > 999)) {
+        if (UNLIKELY(ch != '.')) {
           SET_ERRNO(HPE_INVALID_VERSION);
           goto error;
+          break;
```
indutny: Superfluous break

http_parser.c
```diff
@@ -795,75 +795,49 @@ size_t http_parser_execute (http_parser *parser,
...
         }
 
         parser->http_major = ch - '0';
-        UPDATE_STATE(s_res_http_major);
+        UPDATE_STATE(s_res_http_dot);
         break;
 
-      /* major HTTP version or dot */
-      case s_res_http_major:
+      case s_res_http_dot:
```
indutny: I wonder if state machine for req and req can be united without sacrificing speed? Would branching in `http_minor` be too heavy for us?

http_parser.c
mscdex: Yeah I wasn't sure about it style-wise.

http_parser.c
mscdex: I dunno, I just went for the most straightforward change.

test.c
mscdex: It is over 80 yes, but there are also many other lines in the file that are over 80 so I assumed that wasn't being observed, at least for that file...

indutny: We're moving to 80 column limit in a very very slow way :)

http_parser.c
mscdex: I would say no. It's pretty clear in the RFC that the HTTP version in the request line should be followed by CRLF.

http_parser.c
mscdex: Fixed.

http_parser.c
indutny: Let's remove this comment then.

test.c
mscdex: Changed.

indutny:
Landed in 9f489a4, thank you!

PR #367, opened 2017-05-20
Tolerate non-compliant status line responses

fake666:
- original fix is from daeon: https://github.com/daeon/http-parser/

  "Tolerate web servers which do not return a status message in the
   return response.

   I have noticed this usse on several websites such downloads from
   mediafire.com"

- original pull request: https://github.com/nodejs/http-parser/pull/254
- i merely added the status_cb_called unit test check, there already
  is a test that triggers this without the patch (a 301 without a
  reason phrase).

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -863,10 +863,9 @@ reexecute:
               UPDATE_STATE(s_res_status_start);
               break;
             case CR:
-              UPDATE_STATE(s_res_line_almost_done);
-              break;
             case LF:
-              UPDATE_STATE(s_header_field_start);
+              UPDATE_STATE(s_res_status_start);
+              REEXECUTE();
               break;
             default:
               SET_ERRNO(HPE_INVALID_STATUS);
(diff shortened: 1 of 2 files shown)


http_parser.c
```diff
@@ -915,19 +914,13 @@ size_t http_parser_execute (http_parser *parser,
...
-          UPDATE_STATE(s_header_field_start);
-          break;
-        }
-
         MARK(status);
         UPDATE_STATE(s_res_status);
         parser->index = 0;
+
+        if (ch == CR || ch == LF)
+            REEXECUTE();
```
indutny: Nit: Indent, should be two-spaces less.

fake666:
fixed the indentation, i'll rebase the change once #366 was merged

indutny:
@fake666 thank you for your patience. #366 has just been merged.

fake666:
rebased and good to go :)

indutny:
Landed in 3b0da34d83e4eace29cc49e3b659bf630537b94c, thank you! (Sorry for very long delay)

PR #368, opened 2017-05-24
add: parse query_string and fragment

priest334:


priest334:
add: parse query string and fragment for http request

bnoordhuis:
Thanks, but as this changes the ABI (it makes `http_parser_settings` bigger) we cannot land this until the next major version and that probably won't happen for a long time.

Does `http_parser_parse_url()` work for you in the interim?

priest334:
I just made this changes for my project and didn't consider that aspect. Maybe, I should use alternative method to deal with query parameters. Thanks.

PR #369, opened 2017-06-02
ignore headers transfer_encoding_chunked and content_length to work with http assemblers.

alperakcan:
added ignore_header_transfer_encodig_chunked and ignore_header_content_length to settings.

ignore_header_content_length:
  useful if body length and content length does not match. happens with owa servers. or working with pre http assemblers that unzips compressed content without changing content-length.

ignore_header_transfer_encodig_chunked:
  useful when working with pre-http assemblers which assembles chunked content into one http packet without changing transfer-encoding field.

may be useful for someone.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -1784,8 +1784,9 @@ reexecute:
 
         parser->nread = 0;
 
-        if (settings->ignore_header_content_length == 0) {
-          parser->content_length = 0;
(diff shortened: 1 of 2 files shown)


bnoordhuis:
Thanks, but since it's an ABI change (changes `sizeof(struct http_parser)`), it can't land until the next major release.  That's not expected to happen anytime soon.

Issue #257, opened 2015-07-02
CMake support


wonder-mice:
Hi, I'm going to write CMakeLists.txt for http-parser to use it in my project.
I could put a bit more effort into it (add tests support, install step, etc.) if you are interested in merging CMake support in.
I see that http-parser also has Make and GYP support, and having third build system could be too much.
What do you think about it?

Lupus:
+1

2015-07-02 21:45 GMT+03:00 wonder-mice <notifications@github.com>:

> Hi, I'm going to write CMakeLists.txt for http-parser to use it in my
> project.
> I could put a bit more effort into it (add tests support, install step,
> etc.) if you are interested in merging CMake support in.
> I see that http-parser also has Make and GYP support, and having third
> build system could be too much.
> What do you think about it?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/joyent/http-parser/issues/257>.
>



-- 
Regards,
Konstantin

indutny:
Considering that there not going to be much build changes, I'm totally fine with it. Please go ahead! ;)

Thanks.

nibua-r:
+1

jasnell:
@wonder-mice ... is this still something you're interested in?

RWOverdijk:
:+1:

wonder-mice:
Yes, I have it almost ready, but can't find enough time to groom it before publishing.

rfdickerson:
Similarly, any interest in a pkg-config file? (.pc)?

piribes:
Any news ?

flandr:
Sent out a pull request with a minimal CMake build. It only provides static and shared instances of the main library target, but I think it will be more productive to start basic and iterate from there rather than waiting for a complete port of all the Makefile targets, etc.

PR #374, opened 2017-06-28
Add basic CMake support

flandr:
Only supports building the main library target (static and dynamic).

Refs #257

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,78 @@
+project(http-parser C)
+
+# Required CMake version
+cmake_minimum_required(VERSION 2.8)
+
+set(http-parser_MAJOR_VERSION 2)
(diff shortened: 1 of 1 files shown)


bnoordhuis:
Thanks.  The changes mostly look good to me but I'm curious why you are using set() for everything.  I'm porting node's build system to cmake ([tree](https://github.com/bnoordhuis/io.js/tree/cmake)) and the CMakeLists.txt for http_parser essentially consists out of `add_library(http_parser http_parser.c)` plus a `target_include_directories()` to export the include path.

flandr:
This is just a specialization of a default template I keep around for cmake builds. Some of it is not specifically necessary.

If you are going to do a full port to cmake for node, feel free to ignore this pull request.

PR #377, opened 2017-07-12
[WIP] CMake Support

izaid:
CMake support has been requested a few times, and there has been at least one PR about it. This is a new attempt at it -- I think it is relatively simple. Happy to build it out more if that is what is needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /out/
+build
 core
 tags
 *.o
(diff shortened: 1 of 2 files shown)


izaid:
Commenting on https://github.com/nodejs/http-parser/pull/374/files:

While that PR is more thorough, I believe the style in this PR is more consistent with modern CMake guidelines. We just need to add a couple of more things to cover the entire Makefile option set.

OtakuSenpai:
http://paste.debian.net/990851/

Here's my own cmakelists made in a rush to get the static libs generated which were not with the Makefile.
Feel free to use it and modify it.
I hope cmake support gets added in the near future 👍

PR #391, opened 2017-10-15
Makefile: Relative symlink paths, fixed DESTDIR, and cleaned up uninstall

ShaRose:
This pull request fixes a few issues relating to DESTDIR (Such as allowing you to specify DESTDIR and PREFIX at the same time without overriding eachother), as well as making the symbolic links not use full paths.

It also fixes two issues in uninstall: First, while install used $(LIBNAME), uninstall hard coded libhttp_parser.so. Uninstall also didn't remove all the files: It left a symlink dangling to a file that doesn't exist.


jbergstroem:
LGTM

indutny:
Landed in 4cefc0f and 652b015

indutny:
Thank you!

indutny:
I've reworded the commit message a bit before landing, please consider reviewing the changes just in case.

PR #398, opened 2017-11-29
doc: try and standardize on project name

maclover7:
I went with `http_parser`, but if there's something better, happy to
change it to that.


indutny:
Landed in b11de0f (force push).

PR #407, opened 2018-02-05
parser: add support for SHOUTcast SOURCE method

bnoordhuis:
R=@indutny maybe?

### DIFF
--- a/http_parser.h
+++ b/http_parser.h
@@ -201,6 +201,8 @@ enum http_status
   /* RFC-2068, section 19.6.1.2 */  \
   XX(31, LINK,        LINK)         \
   XX(32, UNLINK,      UNLINK)       \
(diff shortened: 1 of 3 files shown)


http_parser.h
```diff
@@ -201,6 +201,8 @@ enum http_status
   /* RFC-2068, section 19.6.1.2 */  \
   XX(31, LINK,        LINK)         \
   XX(32, UNLINK,      UNLINK)       \
+  /* SHOUTcast */                   \
```
bradisbell: Icecast clients use this method.  Legacy SHOUTcast clients use a completely incompatible protocol (sadly).  The rest of this looks awesome, thanks!

bnoordhuis: I'll update that before I land it, here and in the commit log.

Issue #408, opened 2018-02-05
Heap out-of-bounds read in strtoul (68345541)


Google-Autofuzz:
Hello HTTP Parser team,

As part of our fuzzing efforts at Google, we have identified an issue affecting
HTTP Parser (tested with revision * master b11de0f5c65bcc1b906f85f4df58883b0c133e7b).

To reproduce, we are attaching a Dockerfile which compiles the project with
LLVM, taking advantage of the sanitizers that it offers. More information about
how to use the attached Dockerfile can be found here:
https://docs.docker.com/engine/reference/builder/

TL;DR instructions:
* `mkdir project`
* `cp Dockerfile.HTTP Parser /path/to/project/Dockerfile`
* `docker build --no-cache /path/to/project`
* `docker run --cap-add=SYS_PTRACE -it image_id_from_docker_build`

From another terminal, outside the container:
`docker cp /path/to/attached/reproducer running_container_hostname:/fuzzing/reproducer`
(reference: https://docs.docker.com/engine/reference/commandline/cp/)

And, back inside the container:
`/fuzzing/repro.sh /fuzzing/reproducer`

Alternatively, and depending on the bug, you could use gcc, valgrind or other
instrumentation tools to aid in the investigation. The sanitizer error that we
encountered is here:

```
INFO: Seed: 1729003870
INFO: Loaded 2 modules (621 guards): [0x7f77f35724a0, 0x7f77f3572e50), [0x77ae38, 0x77ae3c), 
/fuzzing/fuzzer: Running 1 inputs 500 time(s) each.
Running: /tmp/poc
ASAN:DEADLYSIGNAL
=================================================================
==11==ERROR: AddressSanitizer: SEGV on unknown address 0x608000010000 (pc 0x7f77f212d2f4 bp 0x000000000000 sp 0x7ffdcc9d3300 T0)
==11==The signal is caused by a READ memory access.
    #0 0x7f77f212d2f3  (/lib/x86_64-linux-gnu/libc.so.6+0x372f3)
    #1 0x7f77f336b5bd in http_parser_parse_url /fuzzing/http-parser/http_parser.c:2371:23
    #2 0x517ae5 in LLVMFuzzerTestOneInput /fuzzing/security-research-pocs/autofuzz/http_parser_parse_url_fuzzer.cc:5:3
[...]

PR #409, opened 2018-02-06
src: fix out-of-bounds read through `strtoul`

indutny:
`strtoul` will attempt to lookup the next digit up until it will stumble
upon an invalid one. However, for an unterminated string as an input
value, this results in out-of-bounds read.

Remove `strtoul` call, and replace it with simple loop.

Fix: #408

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -2367,8 +2367,22 @@ http_parser_parse_url(const char *buf, size_t buflen, int is_connect,
   }
 
   if (u->field_set & (1 << UF_PORT)) {
-    /* Don't bother with endp; we've already validated the string */
-    unsigned long v = strtoul(buf + u->field_data[UF_PORT].off, NULL, 10);
+    uint16_t off;
+    uint16_t len;
+    const char* end;
+    unsigned long v;
+
+    off = u->field_data[UF_PORT].off;
+    len = u->field_data[UF_PORT].len;
+    end = buf + off + len;
+
+    /* NOTE: The characters are already validated and are in the [0-9] range */
+    assert(off + len <= buflen && "Port number overflow");
+    v = 0;
+    for (const char* p = buf + off; p < end; p++) {
+      v *= 10;
+      v += *p - '0';
+    }
 
     /* Ports have a max value of 2^16 */
     if (v > 0xffff) {
--- a/test.c
+++ b/test.c
@@ -3665,6 +3665,30 @@ test_header_cr_no_lf_error (int req)
 }
 
 void
+test_no_overflow_parse_url (void)
+{
+  int rv;
+  struct http_parser_url u;
+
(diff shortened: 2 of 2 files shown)


ploxiln:
👍

http_parser.c
```diff
@@ -2367,8 +2367,22 @@ http_parser_parse_url(const char *buf, size_t buflen, int is_connect,
...
+    unsigned long v;
+
+    off = u->field_data[UF_PORT].off;
+    len = u->field_data[UF_PORT].len;
+    end = buf + off + len;
+
+    /* NOTE: The characters are already validated and are in the [0-9] range */
+    assert(off + len <= buflen && "Port number overflow");
+    v = 0;
+    for (const char* p = buf + off; p < end; p++) {
```
bnoordhuis: C99-ism; not all compilers accept that (because hey, 1999 was only 19 years ago, right?). Can you move the declaration of `p` to the block above?

http_parser.c
```diff
@@ -2367,8 +2367,22 @@ http_parser_parse_url(const char *buf, size_t buflen, int is_connect,
...
+      v *= 10;
+      v += *p - '0';
+    }
 
     /* Ports have a max value of 2^16 */
     if (v > 0xffff) {
```
bnoordhuis: You could move this check inside the loop to protect against `v` wrapping around.

Just a suggestion though; with this PR it's already better than before because errno no longer gets clobbered.

test.c
```diff
@@ -3664,6 +3664,30 @@ test_header_cr_no_lf_error (int req)
...
+{
+  int rv;
+  struct http_parser_url u;
+
+  http_parser_url_init(&u);
+  rv = http_parser_parse_url("http://example.com:8001", 22, 0, &u);
+
+  if (rv != 0) {
+    fprintf(stderr,
+            "\n*** test_no_overflow_parse_url invalid return value =%d\n",
```
bnoordhuis: Teeny tiny nit: no space before `=`, you don't have one below either.

http_parser.c
indutny: Ack.

http_parser.c
indutny: Ack.

test.c
indutny: Ack.

indutny:
@bnoordhuis this is probably not worth re-review, but, just in case, I've pushed `stdlib.h` removal.

bnoordhuis:
:shipit:

indutny:
Landed in 9ce7316, thank you!

bnoordhuis:
@indutny You borked the Fixes link. ;-)

Maybe now is a good time to put out a new release?  The last one was ~18 months ago.

indutny:
Here goes nothing.

PR #422, opened 2018-03-29
Speed up the http_parser_execute loop

brian-pane:
Changes:
* Cache parser->nread in a local variable (same optimization that was
  already in place for parser->state).
* Move the cost of the conditional branch for spaces in tokens out of
  the fast-tokenizer implementation and into the strict-tokenizer
  implementation.

Together, these changes yield a ~15% increase in MB/s and req/s in
the included bench program on x86_64.

Note: this change addresses envoyproxy/envoy issue #[2880](https://github.com/envoyproxy/envoy/issues/2880).

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -49,6 +49,7 @@
 
 #define SET_ERRNO(e)                                                 \
 do {                                                                 \
+  parser->nread = nread;                                             \
   parser->http_errno = (e);                                          \
 } while(0)
 
@@ -56,6 +57,7 @@ do {                                                                 \
 #define UPDATE_STATE(V) p_state = (enum state) (V);
 #define RETURN(V)                                                    \
 do {                                                                 \
+  parser->nread = nread;                                             \
   parser->state = CURRENT_STATE();                                   \
   return (V);                                                        \
 } while (0);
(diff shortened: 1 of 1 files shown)


http_parser.c
```diff
@@ -1236,8 +1239,14 @@ size_t http_parser_execute (http_parser *parser,
             break;
 
           switch (parser->header_state) {
-            case h_general:
+            case h_general: {
+              size_t limit = data + len - p;
+              limit = MIN(limit, HTTP_MAX_HEADER_SIZE);
+              while (p+1 < data + limit && TOKEN(*(p+1))) {
```
bnoordhuis: Can you write this as `TOKEN(p[1])`?  Or perhaps better:
```c
for (++p; p < data + limit && TOKEN(*p); ++p);
```

http_parser.c
```diff
@@ -2404,6 +2415,7 @@ http_parser_pause(http_parser *parser, int paused) {
    */
   if (HTTP_PARSER_ERRNO(parser) == HPE_OK ||
       HTTP_PARSER_ERRNO(parser) == HPE_PAUSED) {
+    uint32_t nread = parser->nread;
```
bnoordhuis: This looks unused.

brian-pane: It's used in the `SET_ERRNO` macro on the next line. I'll add a comment to document that dependency.

derekargueta:
@bnoordhuis ping?

PR #424, opened 2018-04-27
return nonzero on parsing empty url

### LINKED ISSUE #414: `http_parser_parse_url` returns 0 when empty url string is provided
```
#include "http_parser.h"
#include <string.h>
#include <stdio.h>

int main(void) {
    const char* url = "";

    struct http_parser_url parser_url;
    http_parser_url_init(&parser_url);
    int ret = http_parser_parse_url(url, strlen(url), 0, &parser_url);
    fprintf(stdout, "Parse url result: %d\n", ret);

    return 0;
}
```
**Parse url result: 0**
Should `http_parser_parse_url` return non-zero value for error in this situation?

WellingGuzman:
Fixes #414 by returning `1` when the url is empty or `NULL`

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -2303,6 +2303,10 @@ http_parser_parse_url(const char *buf, size_t buflen, int is_connect,
   enum http_parser_url_fields uf, old_uf;
   int found_at = 0;
 
+  if (buf == NULL || buf[0] == '\0') {
+    return 1;
+  }
+
   u->port = u->field_set = 0;
   s = is_connect ? s_req_server_start : s_req_spaces_before_url;
   old_uf = UF_MAX;
--- a/test.c
+++ b/test.c
@@ -3264,6 +3264,18 @@ const struct url_test url_tests[] =
   ,.rv=1 /* s_dead */
   }
 
+  , {.name="empty url"
+    ,.url=""
+    ,.is_connect=0
+    ,.rv=1
+    }
+
+  , {.name="NULL url"
+    ,.url=NULL
+    ,.is_connect=0
+    ,.rv=1
+    }
+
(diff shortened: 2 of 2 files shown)


http_parser.c
```diff
@@ -2303,6 +2303,10 @@ http_parser_parse_url(const char *buf, size_t buflen, int is_connect,
   enum http_parser_url_fields uf, old_uf;
   int found_at = 0;
 
+  if (buf == NULL || buf[0] == '\0') {
```
bnoordhuis: This should be just `if (buflen == 0)`.

The NULL check isn't necessary because `buf` isn't allowed to be NULL when `buflen > 0`, and `buf[0]` is an out-of-bound read when `buflen == 0`.

(It also assumes that `buf` is zero-terminated, which isn't necessarily true.)

WellingGuzman: I was thinking that at first, but I realize as `buflen` is based on `buf`, I should pay more attention to that variable.

In case it was passed was below:
```c
http_parser_parse_url("something", 0, 0, <parser>);
```

This means that this should've be done by checking for `buflen  == 0`? as it will be out of bounds immediately as you said.

If I update the code to:

```c
if (buflen == 0) {
    return 1;
}
```

Could it be merged?

Thanks @bnoordhuis for the info.

WellingGuzman: I made the change locally and the tests passed.

From:

```c
if (buf == NULL || buf[0] == '\0') {
  return 1;
}
```
To:

```c
if (buflen == 0) {
    return 1;
}
```

It makes more sense.

bnoordhuis:
Thanks Welling, landed in 3123273.

PR #428, opened 2018-05-14
GCC 7 does not accept "FALLTHROUGH" as a keyword to omit implicit fall through warning

mloy:
"fall through" is recognized

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -1369,7 +1369,7 @@ reexecute:
           break;
         }
 
-	/* fall through */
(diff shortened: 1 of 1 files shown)


mloy:
Oh, I'm really sorry for that! The second commit should solve this.

bnoordhuis:
Thanks Matthias, landed in 7ba3ee5.

bnoordhuis:
Make that 5b76466, I added a "Fixes:" link.

PR #430, opened 2018-05-29
Chunk extensions callbacks

martin-gabris:
There are already callbacks to get size of chunk and it's body as well as info when chunk was parsed, but I was missing a clean way to get to chunk extension data. Previously, parser just skipped the extension data as long as it started with space or semicolon.

I also changed tests with non-zero chunk extensions to check whether they were parsed correctly.


bnoordhuis:
Yep.  That other PR really shouldn't have landed without a major version bump but that wasn't on my watch.

PR #432, opened 2018-06-05
fix multiple begin message cbs when response starts with CR/LF

mattklein123:
Signed-off-by: Matt Klein <mklein@lyft.com>

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -758,16 +758,14 @@ reexecute:
 
       case s_start_res:
       {
+        if (ch == CR || ch == LF)
+          break;
         parser->flags = 0;
         parser->content_length = ULLONG_MAX;
 
         switch (ch) {
           case 'H':
             UPDATE_STATE(s_res_H);
-            break;
-
-          case CR:
-          case LF:
             break;
 
           default:
--- a/test.c
+++ b/test.c
@@ -2094,6 +2094,7 @@ int
 message_begin_cb (http_parser *p)
 {
   assert(p == &parser);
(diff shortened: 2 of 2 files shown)


mattklein123:
It's not completely clear to me that the RFC allows CR/LF before response start, but this mirrors the logic in the other start states (request and request or response) and I think it's a clear bug to invoke the message begin callback multiple times. Please let me know if you think a different fix would be better.

mattklein123:
@bnoordhuis any chance of getting this looked at this week? Thank you in advance.

http_parser.c
```diff
@@ -766,10 +768,6 @@ size_t http_parser_execute (http_parser *parser,
             UPDATE_STATE(s_res_H);
             break;
 
-          case CR:
-          case LF:
-            break;
-
```
bnoordhuis: Now that the switch has only one case (well, two if you count the default) can you turn it into an if statement?

mattklein123:
@bnoordhuis thanks for the review. Updated.

mattklein123:
@bnoordhuis friendly ping. Can we merge?

mattklein123:
@bnoordhuis friendly ping again. Can we please merge? Thank you!

htuch:
@bnoordhuis any chance we can get a merge here?

mattklein123:
@bnoordhuis @maclover7 do we expect this PR to merge any time soon? If not I will fork http-parser into the envoy proxy org. Thank you!

mattklein123:
Thank you for merging!

PR #452, opened 2018-12-01
Make HTTP_MAX_HEADER_SIZE configurable via gyp

mcollina:
Spin out from https://github.com/nodejs/node/pull/24716.

This also sets the default to 8KB when compiled via gyp.

### DIFF
--- a/http_parser.gyp
+++ b/http_parser.gyp
@@ -47,6 +47,10 @@
     ],
   },
 
+  'variables': {
+    'http_max_header_size%': '8192'
+  },
+
   'targets': [
     {
       'target_name': 'http_parser',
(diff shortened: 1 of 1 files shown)


bnoordhuis (approved):
LGTM % style nit.

http_parser.gyp
```diff
@@ -47,6 +47,10 @@
     ],
   },
 
+  'variables': {
+    'http_max_header_size%': '8192'
```
mcollina: @bnoordhuis do you think we should change this back to 80KB here? We will be passing down the variable anyway in Node.js.

bnoordhuis: I guess it won't hurt backwards compatibility but truthfully, I picked the 80 kb limit at random.

bnoordhuis (approved):

PR #453, opened 2018-12-03
support overriding HTTP_MAX_HEADER_SIZE at runtime

cjihrig:
This commit adds `http_parser_set_max_header_size()`, which can override the compile time `HTTP_MAX_HEADER_SIZE` value.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -25,6 +25,8 @@
 #include <string.h>
 #include <limits.h>
 
+static uint32_t max_header_size = HTTP_MAX_HEADER_SIZE;
+
 #ifndef ULLONG_MAX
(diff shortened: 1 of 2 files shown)


cjihrig:
Is https://ci.nodejs.org/view/All/job/http-parser-test/ the correct CI job? I tried to run it (https://ci.nodejs.org/view/All/job/http-parser-test/26/) with no luck.

mcollina:
It would be good to have a unit test for this here.

bnoordhuis:
A regression test would be nice but `HTTP_MAX_HEADER_SIZE` didn't have one either.

@cjihrig There's no CI job for http-parser.

PR #459, opened 2018-12-19
Disallow empty Content-Length

obatysh:
In https://tools.ietf.org/html/rfc7230#page-30
Content-Length = 1*DIGIT

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -1729,6 +1729,11 @@ reexecute:
             case h_transfer_encoding_chunked:
               parser->flags |= F_CHUNKED;
               break;
+            case h_content_length:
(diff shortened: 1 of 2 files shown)


http_parser.c
```diff
@@ -1729,6 +1729,11 @@ size_t http_parser_execute (http_parser *parser,
             case h_transfer_encoding_chunked:
               parser->flags |= F_CHUNKED;
               break;
+            case h_content_length:
+              // do not allow empty content length
```
bnoordhuis: Please use C89-style comments in this file.

bnoordhuis:
Landed in 3502589. Thanks, Olga!

Issue #435, opened 2018-06-19
Gracefully handle duplicate content-length response headers


derekargueta:
While deploying [Envoy](https://github.com/envoyproxy/envoy), which uses this project for parsing HTTP/1.1 messages, we've seen an occasional bug in a web service where the health-check response contains duplicate `content-length` headers, looking like:
```
< HTTP/1.1 200 OK
--
< Content-Length: 0
< Content-Length: 0
<
```
 This library considers the duplicate `content-length` headers erroneous ([code](https://github.com/nodejs/http-parser/blob/master/http_parser.c#L1414)), and as a result Envoy fails the health-check and won't route traffic to that host. Per [RFC 7230, Section 3.3.2](https://tools.ietf.org/html/rfc7230#section-3.3.2), if the values of duplicate content-length headers are equivalent, it is acceptable to either reject the message or consolidate the values into a single `content-length` header.

This issue is a proposal to implement the latter - if the duplicate content-length headers are found to contain the same value, consolidate them into a single `content-length` header to improve robustness.

[Envoy issue](https://github.com/envoyproxy/envoy/issues/3648)

bnoordhuis:
For background: e2e467b91262246b339fb3d80c8408d498b4a43b - rejecting duplicate Content-Length headers was added as a security measure.

There is no room in `struct http_parser` to store the previous Content-Length value without breaking ABI (relevant when the second header is spread across chunks) so I don't think we could accept identical duplicates even if we wanted to, at least not in v2.x.  And there are no plans for a v3.0 release.

PR #460, opened 2018-12-20
Allow duplicated Content-Length with the same value

obatysh:


indutny:
Hello!

Thank you for submitting this patch. Wouldn't allowing duplicate content-length header be a violation of protocol specification?

obatysh:
RFC https://tools.ietf.org/html/rfc7230#page-31 says that duplicate Content-Length fields with the same decimal value can be either rejected as invalid or accepted by replacing duplicate with a single valid Content-Length field. So, it is not a violation, it is one of the possible options.
Not sure about "replacement" implementation, maybe this pull-request should be improved not to report the Content-Length field second time.

indutny:
Good point!

I'd love to hear opinion of @nodejs/http on that? Would we prefer to reject it or parse it?

ploxiln:
it's worth noting that this costs an additional 8 bytes per `struct http_parser`

bnoordhuis:
Yes. That means it breaks ABI and can't land in a v2.x release. See discussion in #435.

PR #463, opened 2019-03-16
count the URL and headers separately when checking for HPE_HEADER_OVERFLOW errors

caiolrm:
Counting headers, URL and start lines separately because
it was counting everything as headers before the headers_done state,
throwing HPE_HEADER_OVERFLOW errors even when what really caused
the overflow was the request URL being too long

Creating an HPE_URL_OVERFLOW err to make sure embedders
still have a request line limit, but now differentiating what
went over such limit.

### DIFF
--- a/http_parser.h
+++ b/http_parser.h
@@ -63,6 +63,17 @@ typedef unsigned __int64 uint64_t;
 # define HTTP_MAX_HEADER_SIZE (80*1024)
 #endif
 
+/* Maximium uri size allowed. If the macro is not defined
+ * before including this header then the default is used. To
+ * change the maximum uri size, define the macro in the build
+ * environment (e.g. -DHTTP_MAX_URL_SIZE=<value>). To remove
+ * the effective limit on the size of the uri, define the macro
+ * to a very large number (e.g. -DHTTP_MAX_URL_SIZE=0x7fffffff)
+ */
+#ifndef HTTP_MAX_URL_SIZE
+# define HTTP_MAX_URL_SIZE (80*1024)
+#endif
+
 typedef struct http_parser http_parser;
 typedef struct http_parser_settings http_parser_settings;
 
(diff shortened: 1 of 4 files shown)


sam-github:
I think it would be useful to refer to what is being limited with the names from RFC's BNF.

What's the "start line"? Nothing in RFC 2616 is called that.

Request-Line has more than a Request-URI, it also has Method, HTTP-Version, and whitespace, any of which could be unreasonably long when written maliciously.

Does this measure URI completely separately from the other elements of the Request-Line? If so, is there some other limit on the Request-Line size?

http_parser.h
```diff
@@ -160,53 +171,53 @@ enum http_status
...
-  XX(29, PURGE,       PURGE)        \
-  /* CalDAV */                      \
-  XX(30, MKCALENDAR,  MKCALENDAR)   \
-  /* RFC-2068, section 19.6.1.2 */  \
-  XX(31, LINK,        LINK)         \
-  XX(32, UNLINK,      UNLINK)       \
-  /* icecast */                     \
-  XX(33, SOURCE,      SOURCE)       \
+#define HTTP_METHOD_MAP(XX)            \
+  XX(0,  DELETE,      DELETE,  6)      \
```
bnoordhuis: `HTTP_METHOD_MAP` is a public API. Changing it makes it a semver-major change.

caiolrm:
@sam-github 
> What's the "start line"? Nothing in RFC 2616 is called that.

Take a look at the [section 4.1 of RFC 2616](https://tools.ietf.org/html/rfc2616#section-4.1), it briefly mentions it, but there's more detailed info regarding that in RFC 7230 [section 3](https://tools.ietf.org/html/rfc7230#section-3) and [3.1](https://tools.ietf.org/html/rfc7230#section-3.1).

> Request-Line has more than a Request-URI, it also has Method, HTTP-Version, and whitespace, any of which could be unreasonably long when written maliciously.
> 
> Does this measure URI completely separately from the other elements of the Request-Line? If so, is there some other limit on the Request-Line size?

This does measure URI completely separately and you're absolutely correct, there should be a limit for the request line size seeing I changed the way it was validated before. Guess I was too focused on the URI thing and forgot about the rest. Maybe I should add it here:

https://github.com/nodejs/http-parser/blob/3a5a436499bcb8a47f16fc2c3ae999e9ad9eb6d5/http_parser.c#L740-L742

And throw a new error like HPE_START_LINE_OVERFLOW, also have something similar to HTTP_MAX_HEADER_SIZE and HTTP_MAX_URL_SIZE macros like a HTTP_MAX_START_LINE_SIZE. How does that sound?

@bnoordhuis Should I do something like this:

```cpp
#define HTTP_METHOD_LENGTH_MAP(XX) \
  XX(0,  6)                        \
  XX(1,  3)                        \
  XX(2,  4)                        \
  XX(3,  4)                        \
  XX(4,  3)                        \
  XX(5,  7)                        \
  XX(6,  7)                        \
...
```

instead and then have the method_lengths declared using it? Like

```c
static const uint8_t method_lengths[] =
  {
#define XX(num, length) length,
  HTTP_METHOD_LENGTH_MAP(XX)
#undef XX
  };
```

ploxiln:
you may be able to do something like:

```c
static const uint8_t method_lengths[] = 
  {
#define XX(num, name, string) uint8_t(sizeof(#string) - 1),
  HTTP_METHOD_MAP(XX)
#undef XX
  }
```

ploxiln:
(but there may be a slightly less expensive way to just error-out if the request line overall is too big ...)

bnoordhuis:
@ploxiln's suggestion works but it's still possible to exceed the overall limit. This PR stores state in stack locals `url_offset` and `spaces_before_url` but that information is lost when the input is spread over multiple `http_parser_execute()` calls.

caiolrm:
@bnoordhuis true. I can probably do away with `url_offset`, it's just being used for comparison. Gonna try to figure something out for the spaces, though, since currently, we can have multiple before the actual url.

PR #464, opened 2019-03-18
Always skip body for 1xx, 204, and 304 responses

### LINKED ISSUE #251: HEAD, 1xx, 204, 304 have content-length but do not have body
There are some comments in http_parser.cc, line 1837:
        /* Here we call the headers_complete callback. This is somewhat
         * different than other callbacks because if the user returns 1, we
         * will interpret that as saying that this message has no body. This
         * is needed for the annoying case of recieving a response to a HEAD
         * request.
         *
         * We'd like to use CALLBACK_NOTIFY_NOADVANCE() here but we cannot, so
         * we have to simulate it by handling a change in errno below.
         */
Excludes the HEAD request, the 1xx, 204, 304 response also have Content-Length but do not have HTTP body.

So maybe the following code should add:

        if (settings->on_headers_complete) {
          switch (settings->on_headers_complete(parser)) {
            case 0:
             if (status_code / 100 == 1 || status_code == 204 || status_code == 304)
                parser->flags |= F_SKIPBODY;
              break;

azaretsky:
https://tools.ietf.org/html/rfc7230#section-3.3.3:
... any response with a 1xx (Informational), 204 (No Content), or 304 (Not Modified) status code is always terminated by the first empty line after the header fields, regardless of the header fields present in the message, and thus cannot contain a message body.

Fixes: https://github.com/nodejs/http-parser/issues/251

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -886,10 +886,11 @@ reexecute:
       case s_res_status_start:
       {
         /* See RFC 7230 section 3.3.3, step 1 */
-        if (parser->status_code / 100 == 1 || /* 1xx e.g. Continue */
+        if ((parser->status_code >= 100 &&
+             parser->status_code < 200) ||    /* 1xx e.g. Continue */
             parser->status_code == 204 ||     /* No Content */
             parser->status_code == 304) {     /* Not Modified */
-            parser->flags |= F_SKIPBODY;
+          parser->flags |= F_SKIPBODY;
         }
         MARK(status);
         UPDATE_STATE(s_res_status);
--- a/http_parser.c
+++ b/http_parser.c
@@ -474,8 +474,6 @@ static struct {
 };
 #undef HTTP_STRERROR_GEN
 
-int http_message_needs_eof(const http_parser *parser);
-
 /* Our URL parser.
  *
  * This is designed to be shared by http_parser_execute() for URL validation,
@@ -887,6 +885,12 @@ reexecute:
 
       case s_res_status_start:
       {
+        /* See RFC 7230 section 3.3.3, step 1 */
+        if (parser->status_code / 100 == 1 || /* 1xx e.g. Continue */
+            parser->status_code == 204 ||     /* No Content */
+            parser->status_code == 304) {     /* Not Modified */
+            parser->flags |= F_SKIPBODY;
+        }
         MARK(status);
         UPDATE_STATE(s_res_status);
         parser->index = 0;
@@ -1856,7 +1860,7 @@ reexecute:
             /* Content-Length header given and non-zero */
             UPDATE_STATE(s_body_identity);
           } else {
-            if (!http_message_needs_eof(parser)) {
+            if (parser->type == HTTP_REQUEST) {
               /* Assume content-length 0 - read the next */
               UPDATE_STATE(NEW_MESSAGE());
               CALLBACK_NOTIFY(message_complete);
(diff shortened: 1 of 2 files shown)


http_parser.c
```diff
@@ -887,6 +885,12 @@ size_t http_parser_execute (http_parser *parser,
 
       case s_res_status_start:
       {
+        /* See RFC 7230 section 3.3.3, step 1 */
+        if (parser->status_code / 100 == 1 || /* 1xx e.g. Continue */
+            parser->status_code == 204 ||     /* No Content */
+            parser->status_code == 304) {     /* Not Modified */
+            parser->flags |= F_SKIPBODY;
```
bnoordhuis: Indent error, should be two spaces.

Something for a separate commit, since you're here anyway: it might be faster to write `parser->status_code / 100 == 1` as `parser->status_code >= 100 && parser_status_code < 200`.

Long division is moderately expensive and I'm not sure if compilers are smart enough to turn the first into the latter.

test.c
```diff
@@ -1848,8 +1848,7 @@ const struct message responses[] =
   ,.http_minor= 1
   ,.status_code= 101
   ,.response_status= "Switching Protocols"
-  ,.body= "body"
-  ,.upgrade= "proto"
+  ,.upgrade= "bodyproto"
```
bnoordhuis: See #363, it's possible someone might be relying on the current behavior.

bnoordhuis (commented):
While the change LGTM in general I'm kind of worried it might break existing http-parser consumers due to the change in behavior.

test.c
tolikz: This PR only changes the tests for 101 responses. The HTTP_200_RESPONSE_WITH_UPGRADE_HEADER* tests are unaffected (and they all pass).

RFC 7230 explicitly forbids using Content-Length or Transfer-Encoding in 1xx responses ("A server MUST NOT send" etc.) and RFC 2616 also had similarly strict requirements ("All 1xx ... responses MUST NOT include a message-body"), so I think it would probably be ok.

http_parser.c
tolikz: Thanks, corrected.

Quite surprisingly both clang 6.0.0. and gcc 7.3.0 are that smart actually (at least for x86-64) but I changed this comparison anyway.

clang:
```
        addl    $-100, %eax
        cmpl    $100, %eax
        jb      .LBB0_60
```

gcc:
```
        subl    $100, %eax
        cmpw    $99, %ax
        ja      .L170
```

tolikz:
It's quite hard to tell if this behavior change will affect anyone since it seems that these responses are rarely encountered in the wild if at all: 1xx, 204, and 304 responses without a declared message body length are already correctly handled by the parser, while 1xx and 204 with a message body are not legitimate according to the spec, and real servers often do not include content-length or transfer-encoidng in 304 responses (at least IIS, nginx and apache httpd don't).

So there's a great chance that this change is harmless but also purely cosmetic and no-one actually needs it.

bnoordhuis:
cc @indutny - perhaps you can weigh in?

indutny:
I think it might be too hard to propagate description this change to the consumers of http-parser. In order to properly do it, we'll have to do major release.

@nodejs/http : do we consider major release a possibility, or would we prefer to stick with llhttp and do majority of changes there instead?

http_parser.c
tolikz: Wouldn't you mind if I mark this conversation resolved?

bnoordhuis: Sorry, of course! I'll resolve it.

tolikz: Thanks!

tolikz:
Just for the record: I will try to keep this branch rebased on top of the current master in case anyone wants to use it before the next major release.

Issue #468, opened 2019-04-05
http-parser does not correct sanitize NUL characters in header values


htuch:
As described in CVE-2019-9900 and https://github.com/envoyproxy/envoy/issues/6434, http-parser will not reject headers containing NUL in header values (beyond the first character) as required by RFC 7230.

This leads to potentially serious security vulnerabilities for projects, e.g. Envoy that depend on http-parser, as they are likely to expect RFC value sanitization of any headers that are passed. See https://github.com/envoyproxy/envoy/issues/6434 for an example of this and our announcement at https://groups.google.com/forum/#!topic/envoy-announce/VoHfnDqZiAM.

The Node.JS security WG were contacted about this issue on initial discovery.

indutny:
Really sad to hear that!

Would you care to share your test case with us?

I've tried reproducing it from the description, but it appears that ASCII code 0 is not accepted as a part of header value:

```diff
diff --git a/test.c b/test.c
index c3fddd5..af7f7b7 100644
--- a/test.c
+++ b/test.c
@@ -3555,6 +3555,35 @@ test_simple (const char *buf, enum http_errno err_expected)
   test_simple_type(buf, err_expected, HTTP_REQUEST);
 }
 
+
+void
+test_raw_type (const char *buf,
+               size_t len,
+               enum http_errno err_expected,
+               enum http_parser_type type)
+{
+  parser_init(type);
+
+  enum http_errno err;
+
+  parse(buf, len);
+  err = HTTP_PARSER_ERRNO(&parser);
+  parse(NULL, 0);
+
+  /* In strict mode, allow us to pass with an unexpected HPE_STRICT as
+   * long as the caller isn't expecting success.
+   */
+#if HTTP_PARSER_STRICT
+  if (err_expected != err && err_expected != HPE_OK && err != HPE_STRICT) {
+#else
+  if (err_expected != err) {
+#endif
+    fprintf(stderr, "\n*** test_simple expected %s, but saw %s ***\n\n%s\n",
+        http_errno_name(err_expected), http_errno_name(err), buf);
+    abort();
+  }
+}
+
 void
 test_invalid_header_content (int req, const char* str)
 {
@@ -4316,6 +4345,9 @@ main (void)
   test_simple("GET / HTTP/11.1\r\n\r\n", HPE_INVALID_VERSION);
   test_simple("GET / HTTP/1.01\r\n\r\n", HPE_INVALID_VERSION);
 
[...]

PR #469, opened 2019-04-05
http_parser: revert `memchr()` optimization

indutny:
An optimization was introduced in c6097e1d and 0097de58. The crux of
optimization was to skip all characters in header value until either
of CR or LF. Unfortunately, this optimization comes at cost of
inconsistency in header value validation, which might lead to security
issue due to violated expectations in the user code.

Partially revert the optimization, and add additional check to make
general header value parsing consistent.

Fix: #468

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -1496,28 +1496,20 @@ reexecute:
 
           switch (h_state) {
             case h_general:
-            {
-              const char* p_cr;
-              const char* p_lf;
-              size_t limit = data + len - p;
-
-              limit = MIN(limit, max_header_size);
-
-              p_cr = (const char*) memchr(p, CR, limit);
-              p_lf = (const char*) memchr(p, LF, limit);
-              if (p_cr != NULL) {
-                if (p_lf != NULL && p_cr >= p_lf)
-                  p = p_lf;
-                else
-                  p = p_cr;
-              } else if (UNLIKELY(p_lf != NULL)) {
-                p = p_lf;
-              } else {
-                p = data + len;
+              for (; p != data + len; p++) {
+                ch = *p;
+                if (ch == CR || ch == LF) {
+                  --p;
+                  break;
+                }
+                if (!lenient && !IS_HEADER_CHAR(ch)) {
+                  SET_ERRNO(HPE_INVALID_HEADER_TOKEN);
+                  goto error;
+                }
               }
+              if (p == data + len)
               --p;
               break;
-            }
 
             case h_connection:
             case h_transfer_encoding:
--- a/http_parser.c
+++ b/http_parser.c
@@ -1508,7 +1508,7 @@ reexecute:
                 }
               }
               if (p == data + len)
-              --p;
+                --p;
               break;
 
             case h_connection:
--- a/http_parser.c
+++ b/http_parser.c
@@ -1496,20 +1496,24 @@ reexecute:
 
           switch (h_state) {
             case h_general:
-              for (; p != data + len; p++) {
-                ch = *p;
-                if (ch == CR || ch == LF) {
+              {
+                const char* limit = p + MIN(data + len - p, max_header_size);
+
+                for (; p != limit; p++) {
+                  ch = *p;
+                  if (ch == CR || ch == LF) {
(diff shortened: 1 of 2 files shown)


indutny:
cc @nodejs/http .

I do not believe that there is any security risk for Node.js applications. However, since this is causing a public security issue for envoyproxy, I'd suggest we act promptly on this.

Thank you!

test.c
```diff
@@ -4316,6 +4316,9 @@ main (void)
   test_simple("GET / HTTP/11.1\r\n\r\n", HPE_INVALID_VERSION);
   test_simple("GET / HTTP/1.01\r\n\r\n", HPE_INVALID_VERSION);
 
+  test_simple("GET / HTTP/1.0\r\nHello: w\1rld\r\n\r\n", HPE_INVALID_HEADER_TOKEN);
```
htuch: If you could also test with explicit `\0` I think it would be great, since NULs are particularly tricky when it comes to C strings as a special case.

http_parser.c
```diff
@@ -1496,28 +1496,20 @@ size_t http_parser_execute (http_parser *parser,
...
+                if (ch == CR || ch == LF) {
+                  --p;
+                  break;
+                }
+                if (!lenient && !IS_HEADER_CHAR(ch)) {
+                  SET_ERRNO(HPE_INVALID_HEADER_TOKEN);
+                  goto error;
+                }
               }
+              if (p == data + len)
```
htuch: Nit: indent next line

http_parser.c
```diff
@@ -1496,28 +1496,20 @@ size_t http_parser_execute (http_parser *parser,
...
-                else
-                  p = p_cr;
-              } else if (UNLIKELY(p_lf != NULL)) {
-                p = p_lf;
-              } else {
-                p = data + len;
+              for (; p != data + len; p++) {
+                ch = *p;
+                if (ch == CR || ch == LF) {
+                  --p;
```
htuch: If you remove this, can you then skip the `if` below?

test.c
indutny: I agree that there's point to checking `\0` too. However, this works without further changes to the code, because `strlen()` still returns correct value. Perhaps, we should address it in a separate follow-up change?

http_parser.c
indutny: This is just a revert of the change. I'd rather keep it closer to the original code, unless there are other reasons for improving it here?

http_parser.c
```diff
@@ -1496,28 +1496,20 @@ size_t http_parser_execute (http_parser *parser,
...
-              if (p_cr != NULL) {
-                if (p_lf != NULL && p_cr >= p_lf)
-                  p = p_lf;
-                else
-                  p = p_cr;
-              } else if (UNLIKELY(p_lf != NULL)) {
-                p = p_lf;
-              } else {
-                p = data + len;
+              for (; p != data + len; p++) {
```
bnoordhuis: Shouldn't you take `max_header_size` into account here?

indutny: Good point. Will update it in a bit.

indutny: Should be fixed, PTAL.

http_parser.c
htuch: Sure, that's fine.

test.c
htuch: That's fine too.

indutny:
Landed in 2a0d106. Thank you everyone!

@htuch I'd use Node.js security channels for this. https://nodejs.org/en/security/ Thanks again for reporting it!

sam-github:
@indutny @nodejs/http Can anyone confirm that this does not affect Node.js? https://github.com/nodejs/http-parser/pull/469#issuecomment-480421822 --- why would it not affect Node.js?

Because it looks like it could affect Node.js, in which case we'd need a CVE for it and to backport the fix to LTS branches.

indutny:
I believe it would not affect Node.js due to the way the header values are converted to V8 strings in Node. We use `OneByteString(env->isolate(), str_, size_)` and thus `\0` have no special effects on our operation.

PR #480, opened 2019-05-09
Fix http_parser_parse_url to handle very long URLs

piru:
Earlier http_parser_parse_url would incorrectly parse very long URLs: The resulting off and len parameters would just get truncated. Even though very long URLs are typically considered invalid by servers they could still end up being parsed by http_parser_parse_url. Thus it's better handle this situation gracefully.

This change is unfortunately not backwards ABI compatible due to changes in http_parser_url structure field types, hence the major version number bump.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,9 +22,9 @@ PLATFORM ?= $(shell sh -c 'uname -s | tr "[A-Z]" "[a-z]"')
 HELPER ?=
 BINEXT ?=
 SOLIBNAME = libhttp_parser
-SOMAJOR = 2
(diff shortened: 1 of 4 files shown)


piru:
I'd also note that the issue of failing to parse very long URLs properly might also have a security impact: For example if a security critical code would use http_parser_parse_url  function to parse the path of a request and then examine the path to see if it should be passed thru some filter. Due to this issue the parsed and actual paths could differ, which might result in filter bypass.

However, to be exploitable the actual implementation would still been to accept the very long URL.

Issue #97, opened 2012-02-25
Forcing a pause at the end of HTTP headers


AndreLouisCaron:
I've implemented a [simple C++ wrapper](https://github.com/AndreLouisCaron/httpxx) for the `http-parser` library and ran into a problem when using the parser pause feature.

Basically, I have a `Request` object that wraps a `http_parser` instance.  This request object has a `feed()` method which invokes `http_parser_execute()`.  The request object also has a `headers_complete()` which reports the end of HTTP headers.  To make this test reliable (mainly for proxying purposes), I force `http_parser_execute()` to return by pausing the parser calling `http_parser_pause()` from the `on_headers_complete()` callback.  All this seemingly works, except that it doesn't consume the last byte of a request without a body.  To finish processing the request, I have to call `http_parser_execute()` again with the last byte of data.

Now, calling `feed()`/`http_parser_execute()` one extra time is not that much of a problem, except that it makes it prevents clients from accurately finding the exact position of the end of HTTP headers.  In a proxying scenario (HTTP proxy, CGI/SCGI/FastCGI or even WebSockets), where you want to forward the body byte-for-byte, clients end up prefixing the HTTP request body with an extra byte.

This was reported to me on the `httpxx` issue tracker, but I believe it's an issue in `http-parser`.  Basically, the `on_headers_complete()` callback is called upon examining the last byte of the header data, and pausing in that callback prevents the library from marking the last byte (the one the triggered the callback) as consumed.

Visit the `httpx` issue #5 for a [detailed discussion](https://github.com/AndreLouisCaron/httpxx/issues/5).  If it's unclear, I might be able to concoct an [SSCCE](http://sscce.org) that illustrates the problem.

Note: this issue seems related to [pull request 89](https://github.com/joyent/http-parser/pull/89).

jasnell:
@indutny @bnoordhuis ... any reason to keep this open?

vinniefalco:
I'm about to open a new issue. I'd like my C++ wrapper to operate the parser in two stages, first to process the message header as a complete unit, and then to process the body. Returning 1 from `on_headers_complete` is not the answer because you can't resume parsing of the body. So I'm looking through the outstanding issues to see if anyone else seeks similar functionality and I came across this issue.

I'm going to try using the pause feature to implement two stage parsing. Was the bug described fixed? Am I going to encounter a problem with "that last byte?" Is there another way to implement the functionality I desire?

PR #483, opened 2019-06-14
url: treat empty port as default

ethomson:
When parsing URLs, treat an empty port (eg `http://hostname:/`) as if it were unspecified.  RFC 3986 says:

> URI producers and normalizers **SHOULD** omit the port component and its ":" delimiter if port is empty or if its value would be the same as that of the scheme's default.

(Emphasis mine.)  This indicates that URIs MAY be produced with an empty port and the `:` delimiter.

Thus, we stop failing if we end host parsing at the port delimiter.

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -2326,7 +2326,6 @@ http_parse_host(const char * buf, struct http_parser_url *u, int found_at) {
     case s_http_host_v6:
     case s_http_host_v6_zone_start:
     case s_http_host_v6_zone:
-    case s_http_host_port_start:
     case s_http_userinfo:
     case s_http_userinfo_start:
       return 1;
--- a/test.c
+++ b/test.c
@@ -2825,6 +2825,25 @@ const struct url_test url_tests[] =
   ,.rv=0
   }
 
+, {.name="proxy empty port"
+  ,.url="http://hostname:/"
+  ,.is_connect=0
+  ,.u=
+    {.field_set=(1 << UF_SCHEMA) | (1 << UF_HOST) | (1 << UF_PATH)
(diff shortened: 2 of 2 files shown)


indutny:
cc @nodejs/http : are we aware of any possible side effects that might happen in Node.js after landing this?

indutny:
@ethomson thank you for an excellent PR! The change looks great. It has to be considered in the context of Node.js too, however. Let's wait for @nodejs/http to get better understanding of interactions of this PR and Node core.

indutny:
As a side note, I wonder if libgit2 might be interested in using [llhttp](https://github.com/nodejs/llhttp) instead of `http_parser`? The former is faster and verifiable, and is already used in the latest Node.js.

ethomson:
Interesting...!  Thanks for the pointer @indutny.  I’ll take a look, particularly since I’m interested in pausing in my `on_headers_complete` callback.  Would you happen to know offhand if llhttp fixes #97 or if that’s still an issue?

mgorny:
@nodejs/http, ping.

sam-github:
@nodejs/http

http_parser.c
```diff
@@ -2326,7 +2326,6 @@ http_parse_host(const char * buf, struct http_parser_url *u, int found_at) {
     case s_http_host_v6:
     case s_http_host_v6_zone_start:
     case s_http_host_v6_zone:
-    case s_http_host_port_start:
```
bnoordhuis: This changes the behavior of `http_parser_parse_url(is_connect=1)` in that it'll now allow `CONNECT hostname:` as valid input, doesn't it?

PR #484, opened 2019-06-20
Increase default HTTP_MAX_HEADER_SIZE to 16kb

southpolesteve:
Discussion in https://github.com/nodejs/node/issues/27645. I'm not sure it has full sign off yet, but starting the process over here too :)

### DIFF
--- a/http_parser.h
+++ b/http_parser.h
@@ -60,7 +60,7 @@ typedef unsigned __int64 uint64_t;
  * to a very large number (e.g. -DHTTP_MAX_HEADER_SIZE=0x7fffffff)
  */
 #ifndef HTTP_MAX_HEADER_SIZE
-# define HTTP_MAX_HEADER_SIZE (80*1024)
+# define HTTP_MAX_HEADER_SIZE (160*1024)
 #endif
 
 typedef struct http_parser http_parser;


http_parser.h
```diff
@@ -60,7 +60,7 @@ typedef unsigned __int64 uint64_t;
  * to a very large number (e.g. -DHTTP_MAX_HEADER_SIZE=0x7fffffff)
  */
 #ifndef HTTP_MAX_HEADER_SIZE
-# define HTTP_MAX_HEADER_SIZE (80*1024)
+# define HTTP_MAX_HEADER_SIZE (160*1024)
```
ploxiln: this changes the default of 80 KiB to 160 KiB

(nodejs was already setting a different max header size - either using `http_parser_set_max_header_size()`, or not use this http-parser at all, but rather https://github.com/nodejs/llparse - I'm not sure)

southpolesteve: Yep! I see that now. I'll investigate where this is supposed to go. I do wonder if this value should be set to the same as wherever else it is set in node?

ploxiln: This repo is largely for the benefit of other users of this library besides nodejs, I think.

Take a look at:
https://github.com/nodejs/node/blob/v12.4.0/src/node_options.h#L186
https://github.com/nodejs/node/blob/v12.4.0/src/node_options.cc#L580
https://github.com/nodejs/node/blob/v12.4.0/deps/http_parser/http_parser.gyp#L59

note that it looks like this "legacy" parser is not the default, since 12.0 I think:
https://github.com/nodejs/node/blob/v12.4.0/src/node_options.cc#L319-L321

PR #492, opened 2019-09-21
Use -f option when calling ln at install time

ffontaine:
This allows "make install; make install" to work properly.

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@free-electrons.com>
Signed-off-by: Renaud AUBIN <root@renaud.io>
[Retrieved from:
https://git.buildroot.net/buildroot/tree/package/libhttpparser/0001-Use-f-option-when-calling-ln-at-install-time.patch]
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


bnoordhuis:
Thanks, landed in 28f3c35.

PR #497, opened 2020-02-13
Correct test name and numbering

sam-github:
It doesn't matter yet, but two tests had the same name, and same test
position macro.


sam-github:
@indutny Who should merge? I have enough github rights to merge, it appears, but I'm not a regular contributor here, so won't merge without an explicit go-ahead.

indutny:
I don't think we need further review on this PR. Feel free to land it!

Issue #502, opened 2020-03-22
2.9.3 breaks ABI compatibility with 2.9.2 with no corresponding SONAME change


sroracle:
Hello,

After spending quite some time debugging [quassel-webserver](https://github.com/magne4000/quassel-webserver) breaking on my latest system update, I traced the issue back to the upgrade of http-parser from 2.9.2 to 2.9.3 without a corresponding rebuild of node itself. Node wasn't rebuilt because the SONAME of http-parser didn't change.

However, it is clear that the changes introduced in 7d5c99d09f6743b055d53fc3f642746d9801479b horribly broke the ABI of libhttp_parser.so. This is due to the changes made to `enum http_errno` and `struct http_parser` (the changes made to `enum header_states` are also breaking but I'm assuming this is not ABI). The [abi-laboratory report](https://abi-laboratory.pro/index.php?view=compat_report&l=http-parser&v1=2.9.2&v2=2.9.3&obj=eda99&kind=abi) gives a nice summary of this.

The problem I encountered was any request to the app would immediately return a `Bad Request`, but I figure this can manifest in any number of ways - for example, the Fedora people saw that [it made libgit2 segfault](https://bugzilla.redhat.com/show_bug.cgi?id=1814897).

Please consider cutting a new release with an updated SONAME or otherwise making the changes ABI compatible.

Thanks,

Max

bnoordhuis:
Paging @indutny and also @jasnell, @mcollina and @sam-github. 

It seems like a pretty obvious oversight (I caught it immediately while looking at the diff so I'm pretty disappointed no one else did...) so please figure out how to fix this.

mcollina:
Sigh :(.

indutny:
FWIW, I've marked relevant PR as an ABI change. Looks like the details went missing somewhere in the process. Should we cut `2.9.4` with the revert and then `2.10.0`?

Sorry!

jasnell:
As @indutny mentions, this was clearly an abi breaking change. Unfortunate that that was missed in the release process. A revert plus new 2.10.0 should work I think.

bnoordhuis:
> Should we cut 2.9.4 with the revert and then 2.10.0?

Just to be clear, 2.10.0 would contain the ABI change again? With a SONAME bump? What happens when a security vulnerability is found that affects 2.9.4?

I think the change from 7d5c99d can be done in a backwards compatible way. PTAL y'all at #503.

bnoordhuis:
[...]

PR #503, opened 2020-03-23
Fix ABI breakage

bnoordhuis:
Fix ABI breakage introduced in commit 7d5c99d ("Support multi-coding
Transfer-Encoding") by undoing the change in `sizeof(http_parser)`.

Restore the size of the `flags` field and shrink the `index` field from
7 to 5 bits. It track strings up to `strlen("Transfer-Encoding")` bytes
in size so 2^5 == 32 is wide enough.

Fixes: https://github.com/nodejs/http-parser/issues/502

### DIFF
--- a/http_parser.c
+++ b/http_parser.c
@@ -731,6 +731,7 @@ reexecute:
         if (ch == CR || ch == LF)
           break;
         parser->flags = 0;
+        parser->extra_flags = 0;
         parser->content_length = ULLONG_MAX;
 
         if (ch == 'H') {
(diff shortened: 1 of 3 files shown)


sroracle:
This doesn't resolve the problem with `enum http_errno`. I think `INVALID_TRANSFER_ENCODING` needs to be moved to the end of `HTTP_ERRNO_MAP` or something.

bnoordhuis:
Good call, I forgot to update that one. Done in a fix-up commit.

PR #506, opened 2020-04-13
Added a first fuzzer for the parser.

DavidKorczynski:
Hi,
I starting fuzzing http-parser and have created an initial integration with OSS-Fuzz. The fuzzer is in the pull request and the integration with OSS-Fuzz will occur in their own repository (https://github.com/google/oss-fuzz). The build itself occurs within OSS-Fuzz and the code will be continuously fuzzed. If you are interested then please merge the request and provide me with an email to which bug reports will be sent - notice this email will be publicly visible. I am happy to continue creating fuzzers for the project if you would like, but the current fuzzer should be able to reach a decent amount of the code!

David

### DIFF
--- a/fuzzers/fuzz_parser.c
+++ b/fuzzers/fuzz_parser.c
@@ -1,0 +1,37 @@
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+#include "http_parser.h"
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size){
+	// prepare a null-terminated string.
+	char *new_str = (char *)malloc(size+1);
+	if (new_str == NULL){
+		return 0;
+	}
+	memcpy(new_str, data, size);
+	new_str[size] = '\0';
+
+    // http-parser logic
+	static http_parser_settings settings_null = {
+		.on_message_begin = 0
+		,.on_header_field = 0
+		,.on_header_value = 0
+		,.on_url = 0
+		,.on_status = 0
+		,.on_body = 0
+		,.on_headers_complete = 0
+		,.on_message_complete = 0
+		,.on_chunk_header = 0
+		,.on_chunk_complete = 0
+	};
+
+	http_parser parser;
+    http_parser_init(&parser, HTTP_BOTH);
+	http_parser_execute(&parser, &settings_null, new_str, strlen(new_str));
+
+
+	free(new_str);
+	return 0;
+}
+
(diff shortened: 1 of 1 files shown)


fuzzers/fuzz_parser.c
```diff
@@ -0,0 +1,36 @@
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+#include "http_parser.h"
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size){
+    // prepare a null-terminated string.
+    char *new_str = (char *)malloc(size+1);
```
indutny, line 8: Does it have to be a null-terminated string? I don't think the parser cares as long as you provide right `size`.

If it doesn't - is there any point in copying the input?

fuzzers/fuzz_parser.c
```diff
@@ -0,0 +1,36 @@
...
+        ,.on_message_complete = 0
+        ,.on_chunk_header = 0
+        ,.on_chunk_complete = 0
+    };
+
+    http_parser parser;
+    http_parser_init(&parser, HTTP_BOTH);
+    http_parser_execute(&parser, &settings_null, new_str, strlen(new_str));
+
+	free(new_str);
```
indutny, line 33: Nitpick: could you match the code-style of the rest of the repo? In particular: 2-space indent and spaces before braces in function declaration.

fuzzers/fuzz_parser.c
DavidKorczynski: You are right! Fixed.

fuzzers/fuzz_parser.c
DavidKorczynski: Nitpicking is welcomed - should be fixed now.

fuzzers/fuzz_parser.c
```diff
@@ -0,0 +1,26 @@
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+#include "http_parser.h"
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+  static http_parser_settings settings_null = {
```
bnoordhuis, line 8:
```suggestion
  static const http_parser_settings settings_null = {
```

bnoordhuis:
I took the liberty of adding my suggestion. Landed in 7af127d, thanks David!

PR #508, opened 2020-04-28
Added another fuzzer for the urls.

DavidKorczynski:
I noticed we missed a few sections of the code related to URL parsing so I added a fuzzer for this.


bnoordhuis:
Landed in eefbf87, thanks David!

PR #510, opened 2020-04-30
Fix sizeof(http_parser) assert

### LINKED ISSUE #507: armv7hl: Assertion `sizeof(http_parser) == 4 + 4 + 8 + 2 + 2 + 4 + sizeof(void *)' failed
https://github.com/nodejs/http-parser/blob/7af127d3ac497bb036d3ecb4fb24bb5ee7ff4b5b/test.c#L4224

This does not work on armv7hl…

bnoordhuis:
The result should be 32 on both 32 bits and 64 bits architectures
because of struct padding.

Fixes: https://github.com/nodejs/http-parser/issues/507

### DIFF
--- a/test.c
+++ b/test.c
@@ -4343,7 +4343,13 @@ main (void)
   printf("http_parser v%u.%u.%u (0x%06lx)\n", major, minor, patch, version);
 
   printf("sizeof(http_parser) = %u\n", (unsigned int)sizeof(http_parser));
-  assert(sizeof(http_parser) == 4 + 4 + 8 + 2 + 2 + 4 + sizeof(void *));
+
+#if defined(__i386__) || defined(__x86_64__)
+  /* Should be 32 on both 32 bits and 64 bits x86 because of struct padding,
+   * see https://github.com/nodejs/http-parser/issues/507.
+   */
+  assert(sizeof(http_parser) == 32);
+#endif
 
   //// API
   test_preserve_data();


bnoordhuis:
@indutny Can I get a quick LGTM?

mbakke:
@bnoordhuis This works on ARMv7, but fails on i686:

```
$ ./test_g 
http_parser v2.9.4 (0x020904)
sizeof(http_parser) = 28
test_g: test.c:4240: main: Assertion `sizeof(http_parser) == 32' failed.
Aborted
```

indutny (approved):
RSLGTM

thefloweringash:
Turns out the comment from stack overflow I read in https://github.com/nodejs/http-parser/issues/507#issuecomment-621403432 was not sufficient. The actual rule for this case is detailed in [The Lost Art of Structure Packing](http://www.catb.org/esr/structure-packing/#_structure_alignment_and_padding).

> In general, a struct instance will have the alignment of its widest scalar member

Which is not the same as being aligned to the size of the widest member. Apparently x86 and arm differ on the alignment requirements for 64-bit values, where arm requires 8-byte alignment and x86 requires 4-byte alignment. This particular case is covered by a [note about arm and x86 compatibility from intel](https://software.intel.com/content/www/us/en/develop/blogs/understanding-x86-vs-arm-memory-alignment-on-android.html):

> The 8-byte (64-bit) mVar2 results in different layout for TestStruct. This is because ARM requires 8-byte alignment for 64-bit variables like mVar2.

So it seems this test needs to encode the abi requirements.

delroth:
As mentioned by https://github.com/nodejs/http-parser/pull/510#issuecomment-632054042 the fix which was merged here breaks testing on i686-linux (where sizeof == 28, not 32). It seems like the only net effect was trading i686 compat for armv7 compat.

Issue #517, opened 2020-05-21
Request with Transfer-Encoding: chunked and Content-Length is valid per RFC, but rejected with HPE_UNEXPECTED_CONTENT_LENGTH


veshij:
Current implementation rejects requests with `Transfer-Encoding: chunked` and `Content-Length` headers with HPE_UNEXPECTED_CONTENT_LENGTH.
https://github.com/nodejs/http-parser/blob/master/http_parser.c#L1804-L1815

But per https://tools.ietf.org/html/rfc7230#section-3.3.3 that's a valid http request, `Content-Length` must be ignored:
```
       If a message is received with both a Transfer-Encoding and a
       Content-Length header field, the Transfer-Encoding overrides the
       Content-Length. 
```

llhttp parser has the correct behavior: https://github.com/nodejs/llhttp/blob/master/src/native/http.c#L50

indutny:
There is one more sentence right after your quote:

> Such a message might indicate an attempt to
> perform request smuggling (Section 9.5) or response splitting
> (Section 9.4) and ought to be handled as an error.

indutny:
I believe that there is an identical code path in http-parser with regards to handling this.

euroelessar:
One more sentence later it states how this requests must be handled by proxies, which indicates that they should be processed, not outright rejected:
> A sender MUST
> remove the received Content-Length field prior to forwarding such
> a message downstream.

indutny:
😄 

```
ought / ɔt /

   (used to express duty or moral obligation): Every citizen ought to help.
```

So "ought to be handled as an error" at best translates as "SHOULD be handled as an error".

---

There is a paragraph in [RFC 6919](https://tools.ietf.org/html/rfc6919#section-4) as well (even though it is not referenced in RFC 7230):

```
 The phrase "OUGHT TO" conveys an optimistic assertion of an
   implementation behavior that is clearly morally right, and thus does
   not require substantiation.
```

indutny:
I'm inclined to close this issue as it seems that the requirements of protocol specification are very clear on this.

veshij:
Also nginx's header parser: https://github.com/nginx/nginx/blob/76ac67b36f2db9acb2aeb4672f0aeaa8008e7d93/src/http/ngx_http_request.c#L1954-L1962

SaveTheRbtz:
[...]

PR #518, opened 2020-06-03
allow content-length and transfer-encoding: c…

veshij:
fixes https://github.com/nodejs/http-parser/issues/517

### DIFF
--- a/http_parser.h
+++ b/http_parser.h
@@ -303,8 +303,8 @@ struct http_parser {
   unsigned int header_state : 7; /* enum header_state from http_parser.c */
   unsigned int index : 5;        /* index into current matcher */
   unsigned int extra_flags : 1;
+  unsigned int allow_nonrfc_clients : 1;  /* allow headers with both `Content-Length` and `Transfer-Encoding: chunked` set */
   unsigned int lenient_http_headers : 1;
-  unsigned int allow_cl_te: 1;
 
   uint32_t nread;          /* # bytes read in various scenarios */
   uint64_t content_length; /* # bytes in body. `(uint64_t) -1` (all bits one)
--- a/http_parser.h
+++ b/http_parser.h
@@ -302,7 +302,7 @@ struct http_parser {
   unsigned int header_state : 7; /* enum header_state from http_parser.c */
   unsigned int index : 5;        /* index into current matcher */
   unsigned int uses_transfer_encoding : 1; /* Transfer-Encoding header is present */
-  unsigned int allow_length_with_encoding : 1;  /* allow headers with both `Content-Length` and `Transfer-Encoding: chunked` set */
+  unsigned int allow_chunked_length : 1;  /* allow headers with both `Content-Length` and `Transfer-Encoding: chunked` set */
   unsigned int lenient_http_headers : 1;
 
   uint32_t nread;          /* # bytes read in various scenarios */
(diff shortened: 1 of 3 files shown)


veshij:
@indutny what would be the best way to pass new option to the parser? Use settings?

veshij:
Could anyone take a look at the change, please?

http_parser.h
```diff
@@ -304,6 +304,7 @@ struct http_parser {
   unsigned int index : 5;        /* index into current matcher */
   unsigned int extra_flags : 2;
   unsigned int lenient_http_headers : 1;
+  unsigned int allow_cl_te: 1;
```
ploxiln, line 307: All the other fields above add up to 32 bits, this one additional bit pushes into the next 32 bits, and causes the size and offsets of `struct http_parser` to change. That's not good.

You may be able to take one bit from `extra_flags`, it looks like only one of those two bits is actually used, for `F_TRANSFER_ENCODING`. It's awkward because it was intended to support more "overflow" `flags`, but with this change would only support the one. It's also awkward because this is in the "PRIVATE" section of the struct, and library users are intended to set it directly. It looks like there aren't any great options here.

veshij: >It's also awkward because this is in the "PRIVATE" section of the struct, and library users are intended to set it directly. It looks like there aren't any great options here.
Yes, it is. Was trying to make it similar to `lenient_http_headers` in initial implementation.
What do you think if I extend `settings` structure to allow library users to configure behavior there?

veshij: @ploxiln updated PR, could you take a look again, please?

bnoordhuis: Can you move the definition of `allow_cl_te` before `lenient_http_headers`? The position of the latter shouldn't change.

It also needs a comment explaining what it does and `allow_cl_te` feels a bit too cryptic, a more self-descriptive name would be better. Minor style issue: space before `:`

It's okay to reduce `extra_flags` to a 1-bit field but can you rename it to e.g. `uses_transfer_encoding` and update the code accordingly?

The `F_TRANSFER_ENCODING` flag will be unused at that point and it should be safe to remove. It's technically an API break but it's not like you can use it meaningfully now anyway.

test.c
```diff
@@ -4146,6 +4181,9 @@ test_message_pause (const struct message *msg)
   size_t nread;
 
   parser_init(msg->type);
+  if (msg-> allow_cl_te) {
```
bnoordhuis, line 4184:
```suggestion
  if (msg->allow_cl_te) {
```

http_parser.h
veshij: updated the PR

http_parser.h
```diff
@@ -302,7 +301,8 @@ struct http_parser {
   unsigned int state : 7;        /* enum state from http_parser.c */
   unsigned int header_state : 7; /* enum header_state from http_parser.c */
   unsigned int index : 5;        /* index into current matcher */
-  unsigned int extra_flags : 2;
+  unsigned int uses_transfer_encoding : 1; /* Transfer-Encoding header is present */
+  unsigned int allow_nonrfc_clients : 1;  /* allow headers with both `Content-Length` and `Transfer-Encoding: chunked` set */
```
ploxiln, line 305: I think the member name could be a bit better, it seems like it could easily be confused with `lenient_http_headers`. Here's a humble idea for something not too long but more descriptive: `allow_length_with_encoding`

veshij: np, renamed!

ploxiln: ... thinking a bit more, there are other Transfer-Encoding that are not a problem with Content-Length, maybe a better name is `allow_length_chunked` or `allow_chunked_length`

veshij: allow_chunked_length then, updated

ploxiln: looks good to me, thanks

bnoordhuis:
Thanks Oleg, landed with minor indent/line length touch-ups in d0f3b05.

PR #531, opened 2021-04-27
Fix gcc comma at end of enumerator list warning.

paulbartell:
Add dummy "LAST" items to each enum to mask the gcc "comma at end of enumerator list" warning and allow building with -Werror.


viccpp:
Trailing comma is permitted [since C99](https://en.cppreference.com/w/c/language/enum) (and C++11)

paulbartell:
@viccpp . Very true. The FreeRTOS [coreHTTP](https://github.com/FreeRTOS/coreHTTP) library that depends on http-parser targets C90 compilers.

